PR #14, opened 2018-07-05
Codacy items and Cleanup

Baklap4:
# Changes in the pull request
* Fixed loads of Codacy issues, some highlights:
    * Make use of overloading instead of defaulting values.
    * Added static keywords to classes which were completely static
    * Braces
    * Default cases + Custom exception.
* Removed unused / dead code as per Contributing guidelines: https://github.com/UltraStar-Deluxe/Play/blob/master/CONTRIBUTING.md
* Removed Todo's
    * Todo's mainly will clutter code and should be fixed within your feature/bugfix/hotfix branch.
    * If a todo cannot be solved earlier an issue should be made to track this status.
* Added some xml doc to methods if it had a comment already.
    * Would be nice to XML document every method, class, enum and interface even if the instance is private.
    * This will prevent your code being cluttered with inline comments.
        * If an inline comment is needed your method is probably doing too much. Keep it simple, stupid.
* Fixed a test to actually test something.


basisbit:
> Removed unused / dead code as per Contributing guidelines: https://github.com/UltraStar-Deluxe/Play/blob/master/CONTRIBUTING.md

... not sure if removing TODOs and still to be implemented stuff is the right way to go, but it would be exactly what the contributing.md says 😸

Baklap4:
Gheheh well personally i hate just empty classes which don't do anything.
If you want to create something you'll do it then and finish it (on a feature branch) ;)

As for the parameter/variables being called `value` the ones i saw i changed to a proper name as `value` is a contextual keyword. As for getters/setters i'd approve of this however in normal methods i wouldn't.

What would you guys like?

basisbit:
yes, please we should definitely use meaningful names for everything (except maybe getters/setters if it is definitely obvious)

PR #22, opened 2018-07-31
fix some compilation issues

barbeque-squared:
### What does this PR do?
It fixes some compile issues. The testcase will probably fail but I just have no idea what it is supposed to test anyway (if I were to test ClearPlayers, I'd expect it to assert that GetPlayers returns an empty list or something (which also implicitly checks for it being an actual object) but maybe it has some historic value, so I've not touched that for right now)

### Additional Notes
I am on 2018.2.0f2 (instead of 2018.1.4f1) but I think that's not a big deal?

### DIFF
--- a/UltraStar Play/Assets/Editor/ManagersTests.cs
+++ b/UltraStar Play/Assets/Editor/ManagersTests.cs
@@ -10,9 +10,7 @@ public class ManagersTests
     public void TestPlayersManagerCreation()
     {
         PlayersManager.ClearPlayers();
-        //~ PlayersManager playersManager;
-        //~ Assert.IsNotNull(playersManager);
-        Assert.IsNotNull(null);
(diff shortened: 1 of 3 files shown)


UltraStar Play/Assets/Editor/ManagersTests.cs
```diff
@@ -10,8 +10,9 @@ public class ManagersTests
     public void TestPlayersManagerCreation()
     {
         PlayersManager.ClearPlayers();
-        PlayersManager playersManager = new PlayersManager();
-        Assert.IsNotNull(playersManager);
+        //~ PlayersManager playersManager;
+        //~ Assert.IsNotNull(playersManager);
+        Assert.IsNotNull(null);
```
frankkienl: This assertion will always fail. Maybe comment this line too?

PR #23, opened 2018-08-01
Add iOS support

frankkienl:
### What does this PR do?

Add iOS support.
Added 'microphoneUsageDescription', this is needed to inform the user why microphone access is needed. (App will crash when microphone is accessed if there is no description)

### Closes Issue(s)

none

### Motivation

I wanted to make sure it builds and runs on iOS

### More

Buttons appear a bit small. But the UI is still a work in progress. Let's look at this later.

### Additional Notes

When building for iOS, make sure to select a 'Development Team' in XCode, otherwise the build will fail. A personal development team (any Apple ID) will do. No need to sign up for the Apple AppStore subscription.

### DIFF
--- a/UltraStar Play/ProjectSettings/ProjectSettings.asset
+++ b/UltraStar Play/ProjectSettings/ProjectSettings.asset
@@ -149,7 +149,8 @@ PlayerSettings:
   resolutionScalingMode: 0
   androidSupportedAspectRatio: 1
   androidMaxAspectRatio: 2.1
-  applicationIdentifier: {}
+  applicationIdentifier:
+    iOS: org.ultrastardx.play.alpha
   buildNumber: {}
   AndroidBundleVersionCode: 1
   AndroidMinSdkVersion: 16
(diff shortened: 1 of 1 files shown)


UltraStar Play/ProjectSettings/ProjectSettings.asset
```diff
@@ -149,7 +149,8 @@ PlayerSettings:
   resolutionScalingMode: 0
   androidSupportedAspectRatio: 1
   androidMaxAspectRatio: 2.1
-  applicationIdentifier: {}
+  applicationIdentifier:
+    iOS: org.ultrastardx.play.alpha
```
basisbit: I guess it is only ultrastar instead of ultrastardx, according to the games name. Also, is it the usual thing to do to add the version in the app identifier? Wouldn't that break any possible updating?

UltraStar Play/ProjectSettings/ProjectSettings.asset
```diff
@@ -287,7 +385,7 @@ PlayerSettings:
   enableCrashReportAPI: 0
   cameraUsageDescription: 
   locationUsageDescription: 
-  microphoneUsageDescription: 
+  microphoneUsageDescription: This game needs access to the microphone. Highscores will be awarded for correctly singing along with the music.
```
basisbit: should probably be changed to `along "to" the music`

frankkienl: 👍

UltraStar Play/ProjectSettings/ProjectSettings.asset
frankkienl: For me at least, and only with iOS, yes. (For Android it's not that big of a deal)

It's the usual thing to not use the final app identifier before going to production.
And it's correct that it will break updating. But that's by design. That way the production app won't have to deal with old data from non-released version.

And although we won't have multiple supported versions (there will be only 1 master)
Its better to do pre-release build with a non-final bundle id.

And that's because of Apple.
When you build an iOS app that bundle-id is linked to your Apple account. Meaning your account will be the only account that can every build with that bundle id. 
I figured it would not be wise to link the final the bundle id to my account.

The final bundle id should be linked to your Apple account, as maintainer of this repo.

UltraStar Play/ProjectSettings/ProjectSettings.asset
basisbit: @frankkienl could you please update your branch/fork and thus also update this pull request with this small change? Thanks!

basisbit: @frankkienl ping

frankkienl: I would recommend to close this pull request. 

(Don't worry about the changes, they're all automatically generated when you select iOS target in Unity.)

We can work on iOS support later, when we've talked about how we will do deployment.

basisbit: @frankkienl maybe we can talk about this next weekend in Utrecht?

frankkienl: Perfect

achimmihca:
Can we close this PR?

PR #24, opened 2018-08-05
documentation: s_ for static members

barbeque-squared:
I was looking at the `src/model/*Manager.cs` and they have all these static class members that start with `s_`. Might as well add it in the documentation.


basisbit:
wasn't sure if we should stick with the s_ prefix. Would like to hear some comments or a small discussion if we want to stick with these prefixes.

basisbit:
after having learned quite a bit about coding style guides and best practices, I'd like to to back to full default Visual Studio / unity naming conventions, to make it easier for others to participate in this project. @barbeque-squared would this be ok for you?

PR #25, opened 2018-09-25
Song parser

barbeque-squared:
### What does this PR do?
* Adds a SongBuilder object to make parsing life easier
* Adds a Voice object, which is basically m_voicesSentences but also has the name
* It also changes a stray tab to spaces somewhere, which isn't really related to this PR but I was working like 1 line above it anyway.
* I didn't add the .meta files because I use a different Unity version. Either someone with a correct version should add them, or just gitignore them.
* I couldn't find anything regarding rap notes, but they weren't there before either.

I might have accidentally branched from my other PR instead of master, whoops.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,7 +33,7 @@ This project just started! Please wait for the first few versions to be released
 - only strictly typed variables
 - camelCase: parameter, 
 - PascalCase: Method, Class, EEnum, NameSpace, Property, IInterface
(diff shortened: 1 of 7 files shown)


basisbit:
this pr includes #24 , thus there would be merging conflicts. Please try to keep different topics of changes separated in different pull requests, (or all in one, but with separate issues for discussing these changes)

barbeque-squared:
I'll redo the changes on a clean branch, create a new PR for that one and then close this one.

barbeque-squared:
Closed now that #26 exists.

PR #26, opened 2018-10-06
Song parser 2

barbeque-squared:
### What does this PR do?
* This PR is the same as #25 but with the correct branching
* Adds a SongBuilder object to make parsing life easier
* Adds a Voice object, which is basically m_voicesSentences but also has the player name
* It also changes a stray tab to spaces somewhere, which isn't really related to this PR but I was working like 1 line above it anyway.
* I didn't add the .meta files because I use a different Unity version. Either someone with a correct version should add them, or just gitignore them.
* I couldn't find anything regarding rap notes, but they weren't there before either.
* It will not shift overlapping notes/sentences anymore, perhaps it's better to do that in a separate function after the parsing is done.
* The specific beat at which a linebreak occurs is not saved anywhere (though I don't think it did that to begin with). It's probably better to just make some 'algorithm' to compute the optimal linebreak beat between two sentences at a later stage? In USDX linebreaks happen when the file tells it to, which sounds nice but is completely inconsistent between songs that have long breaks in them: one will have the previous sentence on it for ages, another song the next sentence. It's also annoying in the USDX editor, but I'll make a separate issue for this.

### DIFF
--- a/UltraStar Play/Assets/src/model/song/Song.cs
+++ b/UltraStar Play/Assets/src/model/song/Song.cs
@@ -8,10 +8,10 @@ public class Song
 {
     private readonly string m_path;
     private readonly string m_folderPath;
-    private readonly List<List<Sentence>> m_voicesSentences;
+    private readonly List<Voice> m_voices;
     private readonly Dictionary<ESongHeader, System.Object> m_headers;
 
-    public Song(Dictionary<ESongHeader, System.Object> headers, List<List<Sentence>> voicesSentences, string path)
+    public Song(Dictionary<ESongHeader, System.Object> headers, List<Voice> voices, string path)
     {
         if(headers == null || headers.Count == 0)
         {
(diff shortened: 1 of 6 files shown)


basisbit:
rap notes are notes where the player gets points as long as an note is registered. this is similar to how Sony implemented that in SingStar for PlayStation 3/4

barbeque-squared:
Apparently, codacy doesn't give me _all_ (new) issues in a file immediately, so I didn't see that there were still some problems last week. (offtopic: codacy is really damned slow, and when it's done there's still caching going on so it basically takes me like 1 minute to fix all the codacy complaints and push them, then 5 minutes of waiting for the next batch of complaints. This gets annoying really fast)

Anyway, codacy is happy now, and I also came across a compiler warning that I hadn't seen before so I fixed that too. Tests are missing, but I have no idea where those are [supposed to go] in the first place.

> rap notes are notes where the player gets points as long as an note is registered. this is similar to how Sony implemented that in SingStar for PlayStation 3/4

Yes, I figured that, but how do I recognize Rap and RapGolden notes in a notes.txt file? `:` = Normal, `*` = Golden and `F` = Freestyle, but what letters are the other two? It wasn't in the implementation before and I can't find it anywhere online either.

basisbit:
- ':' Normal (1x points)
- 'F' Freestyle (no points)
- '*' Golden (2x points)
- 'R' Rap (1x points)
- 'G' Golden Rap (2x points)

Regarding codacy being to slow, I suggest you to use some lint plugin like sonarLint or similar. Thus you'll see possible issues already while coding.

Issue #1, opened 2018-06-19
set up automatic builds


basisbit:
- [x] automatic build for each commit & pr
- [x] github integration
- [x] require successful build before merging
- [x] build logs publicly available
- [x] use same code / script for local builds where possible
- [x] upgrade travis to be used from a webhook or as a Github App
- [x] upload windows travis build artifacts upon commit to master.

Baklap4:
Another needed thing is to cache `Unity-2018.1.4f1` unless a newer version is found remote. This will decrease the build time.

basisbit:
build currently fails because with unity 2018.1 it now is impossible to activate the unity install using a personal license. There exists a workaround: create and use a pre-activated docker image

basisbit:
others already worked on figuring out how to use unity + docker in travis-ci with personal license: http://www.robinryf.com/blog/2017/09/30/running-unity-inside-docker.html

Baklap4:
Travis should also be updated to be spinned up by using webhooks or installed as an app. This is because services (as it's now configured) are getting deprecated on the **first of october.**. 

> GitHub plans on shutting down GitHub Services so we can shift focus to other areas of the API, such as strengthening GitHub Apps and GraphQL, and improving webhooks. The intention of GitHub Services was to allow third-party developers to submit code for integrating with their services, but this functionality has been superseded by GitHub Apps and webhooks. GitHub Services has not supported new features since April 25, 2016, and we plan to officially deprecate it on October 1st, 2018.

![image](https://user-images.githubusercontent.com/3918517/42352082-a95fba3c-80b8-11e8-9122-88263d34efed.png)

Baklap4:
Upload travis artifacts (windows, osx, linux) to github upon tag creation would be enough to setup our first kind of continuous deployment

basisbit:
@achimmihca for your information: TravisCI automated build-check and-testcase-execution is finally working again. I'll re-enable passing the TravisCI-build check integration as requirement before merging a pull-request, if you are ok with that?

achimmihca:
Nice to hear that!
Does the automatic build also create an executable? I mean, is the Unity build started for some target platform?

It would be nice to have a link to the UltraStar Play nightly version for windows / linux / mac. Note that this nightly should also include needed game resources such as I18N files.

basisbit:
Build of each master branch commit now finally works! here is a build of 43c0d8b:
[...]

Issue #28, opened 2018-10-17
Switch to .NET 4.x


nikolaosginos:
I would propose targeting .NET 4.x for this project to be able to use the Task Parallel Library for easier threading implementation the usage of newer C# language features.

I have already done this in my local clone. I could submit a pull request.

Was there any specific reason for staying with 3.5 or was this just the default setting?

basisbit:
we are using unity 2018.1.4f1, which still has some known issues with .Net 4.x. If you get unity to compile this project and run tests on travis-ci or a docker container, we'll gladly upgrade to 2018.2.x

basisbit:
with current versions of Unity properly supporting .Net 4.x and travis-ci now using a version that also supports .net 4, there should be no reason left not to switch now. @nikolaosginos this was your suggestion, thus could you please create the pull request for this change?

nikolaosginos:
> with current versions of Unity properly supporting .Net 4.x and travis-ci now using a version that also supports .net 4, there should be no reason left not to switch now. @nikolaosginos this was your suggestion, thus could you please create the pull request for this change?

Yes, I will take care of this

PR #32, opened 2018-11-01
attempt at fixing travis-ci build and test checks

basisbit:
### What does this PR do?
Hopefully fixes the travis-ci autobuild/-test issues.

### Related Issue(s)

this is part of #1 

### Motivation

Automatic building and reporting simplifies developers' and project maintainers' life by reducing workload.

### More

- [ ] please do not merge yet, this is still work in progress


basisbit:
Travis-CI is finally green 🥂

.travis.yml
```diff
@@ -1,11 +1,14 @@
...
-language: generic
-
-# use a Mac OSX image
-os: osx
-
+sudo: required
+language: bash
+env:
+  global:
+  - IMAGE_NAME=gableroux/unity3d:2018.2.6f1
```
barbeque-squared: Contrary to what https://github.com/UltraStar-Deluxe/Play/wiki/Compiling-the-game and #28 say, this uses a more recent Unity version. Wiki edits are probably separate from the codebase, and I'm guessing the newer version is deliberate here, but don't forget to update the PR and wiki after merging.

UltraStar Play/Assets/Editor/BuildCommand.cs
```diff
@@ -0,0 +1,108 @@
+﻿using UnityEditor;
```
barbeque-squared: This file has mixed tabs and spaces (I'm surprised that Codacy didn't pick up on it?). Should be space-only.

tools/travis/before_script.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+set -e
+set -x
+mkdir -p /root/.cache/unity3d
+mkdir -p /root/.local/share/unity3d/Unity/
```
barbeque-squared: One mkdir ends without a trailing slash, the other with. I know that both do exactly the same, though I usually try to keep them consistent within the same file, especially if they're two consecutive lines.

tools/travis/build.sh
```diff
@@ -1,57 +1,36 @@
...
+  echo "Run succeeded, no failures occurred";
+elif [ $UNITY_EXIT_CODE -eq 2 ]; then
+  echo "Run succeeded, some tests failed";
+elif [ $UNITY_EXIT_CODE -eq 3 ]; then
+  echo "Run failure (other failure)";
+else
+  echo "Unexpected exit code $UNITY_EXIT_CODE";
+fi
+
+ls -la "UltraStar Play/$BUILD_PATH"
```
barbeque-squared: Is the output of this ls used somewhere? In understand why the echo's are here, just not what this ls is doing here.

tools/travis/activat.ion
```diff
@@ -0,0 +1,36 @@
...
+            <Feature Value="19"/>
+            <Feature Value="62"/>
+        </Features>
+        <DeveloperData Value="AQAAAEY0LVI4VEstNU1SOS1UVUNVLUtXOFgtWUM1Mg=="/>
+        <SerialMasked Value="F4-R8TK-5MR9-TUCU-KW8X-XXXX"/>
+        <StartDate Value="2017-12-16T14:55:19"/>
+        <UpdateDate Value="2018-11-02T23:05:29"/>
+        <InitialActivationDate Value="2017-12-16T14:55:19"/>
+        <LicenseVersion Value="6.x"/>
+        <ClientProvidedVersion Value="2018.2.6f1"/>
```
barbeque-squared: See the other comment about the Unity version.

tools/travis/build.sh
basisbit: the folder output is only there to simplify analysis if travis-ci build didn't work as intended or if if says "green" but actually should report "red"

tools/travis/activat.ion
basisbit: Wasn't able to get the older versions to work with travis-ci. It supports the desired testing + compile checks for CI purposes of the 2018.1.4f1 version project files. Also, see #28

UltraStar Play/Assets/Editor/BuildCommand.cs
basisbit: sorry, my fault. Neither the compiler nor most modern IDEs care about tabs or spaces at line beginning, so I didn't notice the difference. I'll update the PR.

barbeque-squared: Looks good now.

tools/travis/build.sh
barbeque-squared: Had a feeling it was something like that (but had no way to be sure), so no problem leaving it in.

tools/travis/activat.ion
barbeque-squared: Green light for using it then. I use 2018.2.0f2 myself, and as long as I don't commit anything involving .meta files, I haven't noticed any issues on the current .NET.

PR #45, opened 2019-10-09
Theme demo

achimmihca:
I implemented a demo for what I had in mind with respect to Theming.
There is a ThemeManager along with ThemeableImage, ThemeableAudio and ThemeableColor scripts that can be dropped at any GameObject to change Image.sprite, AudioClip and Image.color values respectively.

These resources are loaded dynamically from Unity3D's Resources folder inside of Assets. A theme can have a parent. The parent's resources will be used as fallback (similar to how WordPress base themes look for resources).
Sprites and AudioClip are assets that are loaded directly by Unity. The colors are declared in an xml file, similar to how Android specifies colors in resources files.

I wanted to avoid the brittle mess of managing a ton of strings that point to files and colors in the resources folder. These things cannot be refactored and checked for existance by Unity.
Instead, the resources are specified using enums. Of course this also leads to an overhead because the enums have to be maintained when adding / changing resources.
Still, I think this should be fine for now.
One could implement a check that the values for the enums really point to existing files.

Note that this is intended as a first idea to show how theming can be done in Unity. It's a general solution and not limited to UltraStar Play.


basisbit:
I really like the extensive use of enums!

PR #48, opened 2019-10-18
new song select with roulette design of USDX

achimmihca:
see title

The main things missing now to actually use USDP are
- some scenes for configuration of important options (e.g. players, game resolution, and mics)
- real-time and accurate pitch detection (which probably requires an external lib)


UltraStar Play/Assets/Common/Util/ImageManager.cs
```diff
@@ -0,0 +1,29 @@
+
+using System.Collections.Generic;
+using System.IO;
+using UnityEngine;
+
+public class ImageManager
+{
+
+    private static readonly Dictionary<string, Sprite> spriteCache = new Dictionary<string, Sprite>();
```
basisbit: This Dictionary should probably be changed to be size-limited and use for example least frequently used as cache replacement policy, once the available application memory gets low. If this cache is used also for song cover images, it can get quite big depending on how many songs there are and depending on the image format and size.
Could you please change this to [MemoryCache ](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.caching.memorycache?redirectedfrom=MSDN&view=netframework-4.8)or similar?

UltraStar Play/Assets/Common/Util/GameObjectUtils.cs
```diff
@@ -21,4 +24,29 @@ public static T FindComponentWithTag<T>(string tag)
...
+        {
+            foreach (Transform rootTransform in rootTransforms)
+            {
+                T obj = rootTransform.GetComponentInChildren<T>(true);
+                if (obj != null)
+                {
+                    return obj;
+                }
+            }
+            return null;
```
basisbit: Please write a log entry for these cases and include stacktrace. Maybe this would be a good place to throw an exception.

PR #49, opened 2019-10-21
Some essential parsing fixes without any tests

barbeque-squared:
### What does this PR do?

Parse more things that are probably okay to parse. It's mostly so that it doesn't completely break on a significant portion of currently already existing songs. In detail, it fixes:
- support for the '- beat1 beat2' linebreak format (or more specifically, it just ignores everything after the first beat. I know it's a 'feature' in USDX.
- adds some overflow exception catches to int parsing

_Yes_ it is super much behind master at this point, and _yes_ it also does not have any tests. I'm only making this pull request because in the past four months, I've just not found the time to make it a proper pull request. It should be fairly trivial to a) decide if you want to support these things at all and b) reimplement them (with proper tests) in the current master.

Maybe it's already in the current master, if so you can just close this PR.

### Motivation

While fixing songs from the MLK project, I discovered some missing 'features' that we might or might not want to support.

### DIFF
--- a/UltraStar Play/Assets/src/model/song/builder/SongMetaBuilder.cs
+++ b/UltraStar Play/Assets/src/model/song/builder/SongMetaBuilder.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Text;
+using System.Text.RegularExpressions;
 
 static class SongMetaBuilder
 {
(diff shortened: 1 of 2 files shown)


UltraStar Play/Assets/src/model/song/builder/SongMetaBuilder.cs
```diff
@@ -63,15 +64,15 @@ public static SongMeta ParseFile(string path, Encoding enc = null)
                 {
                     requiredFields[tag] = val;
                 }
-                else if (tag.StartsWith("p", StringComparison.Ordinal) && tag.Length > 1)
+                else if (tag.StartsWith("p", StringComparison.Ordinal) && tag.Length > 1 && Regex.IsMatch(tag, @"^p\d+$"))
```
basisbit: Please change such often called Regex checks to a compiled regex, which is instantiated once and then reused.
(for example Regex myRegex = new Regex(pattern, RegexOptions.Singleline | RegexOptions.Compiled);

achimmihca:
I changed the parsing of song files a bit a while ago. Thereby I considered the `- beat1 beat2` format for linebreaks.

achimmihca:
I just implemented the other topics from this pull-request in #118 .
So I will close this PR now.

PR #51, opened 2019-10-22
improve note recording

achimmihca:
I tried to improve the pitch detection.
Now I am pretty sure that I used the pitch detection lib correctly but it is still too slow and inaccurate for proper gameplay.

However, using the PerfectSinger script, which records all notes at their correct pitch, I discovered and fixed some issues in PlayerNoteRecorder.

Besides this, there are only small code improvements.


UltraStar Play/Assets/Common/Graphics/ImageManager.cs
```diff
@@ -9,7 +9,7 @@ public class ImageManager
 {
     // When the cache has reached the critical size, then unused sprites are searched in the scene
     // and removed from memory.
-    private static readonly int criticalCacheSize = 50;
+    private static readonly int criticalCacheSize = 8;
```
basisbit: how did you come up with this number? That seems a bit low for some scenes with for example 6 different player avatar sprites, golden notes sprite, normal notes sprite, various theme elements,...

UltraStar Play/Assets/Common/Graphics/ImageManager.cs
```diff
@@ -59,11 +59,16 @@ private static Sprite CreateNewSprite(string path)
...
-        Image[] imagesInScene = GameObject.FindObjectsOfType<Image>();
-        foreach (Image image in imagesInScene)
+        // Iterate over all sprites in the scene that are referenced by an Image or ISpriteHolder
+        // and remember them as still in use.
+        foreach (Transform transform in GameObject.FindObjectsOfType<Transform>())
         {
-            usedSprites.Add(image.sprite);
+            Image image = transform.GetComponent<Image>();
+            ISpriteHolder spriteHolder = transform.GetComponent<ISpriteHolder>();
+            if (image != null || spriteHolder != null)
```
basisbit: shouldn't this be `if (image != null && spriteHolder != null)`?

UltraStar Play/Assets/Common/Graphics/ImageManager.cs
achimmihca: Ups, that one was only for testing that it works. I have 9 song for testing, so when the 9th cover is loaded, the unused resources are removed and I wanted to trigger this.
Will fix this with the next PR.

achimmihca:
> Player- and microphone setup probably are more crucial right now to be able to release a first alpha version.
The thing is that the way microphones are handled can depend on the library that is used. I don't know if the device name that is used by Unity to identify mics is always possible when working with libs (although probably yes when the name is coming from the OS).

Anyway, I will continue with what we got so far.

PR #53, opened 2019-10-29
LeanTween

achimmihca:
I added and used LeanTween to re-implement some small animations.
I really love this lib!

Also: some minor improvements of the Theme Demo and a fix of the ImageManager cache size.


basisbit:
personally, I am not too happy with having LeanTween plus all of its documentation and so on in this repository. Please use git submodules for that.

achimmihca:
> personally, I am not too happy with having LeanTween plus all of its documentation and so on in this repository. Please use git submodules for that.

Me neither, especially when looking at the other libs that I want to use. They would blow up our repo.

I thought about cloning the LeanTween repo on GitHub as submodule.
However, Unity requires the scripts in our Assets folder. Cloning only the Assets folder [that is relevant for](https://github.com/dentedpixel/LeanTween/tree/master/Assets/LeanTween) us from LeanTween seems [not possible](https://stackoverflow.com/questions/3866733/how-to-clone-only-a-folder-from-a-git-submodule).

A workaround could be to download the LeanTween repo as submodule and create a symlink inside our Assets folder to the relevant part of the LeanTween repo. But I don't think it will work well when people use different operating systems (symlinks of linux vs windows). And I don't know if Unity will handle this well.

My other idea was to clone the repo (maybe temporarily), copy the relevant folders to the needed places, and add a gitignore such that they are not included in our repo. Cloning and copying could be done via script.
Would this be ok for you? Or do you know a better approach?

Unity is really missing a proper dependency / package management and build tool like gradle.

basisbit:
How about we create another repo in https://github.com/UltraStar-Deluxe/... for UltraStar Play dependencies and within that repository we keep the dependencies / folders that are required and which then can be added as submodule into the Assets folder? This would at least allow for a basic layer of separation between dependencies and our own code.

achimmihca:
I think there must be a better way.
Maybe some git command can do the trick (I read about --filter options, git subtree, sparse-checkout, ...).
I suggest to postpone this until we took a closer look at git.

Anyway, the goal is clear: we don't want the dependencies to be in the repo of UltraStar Play.
Thus, I will add a gitignore for now. What remains is the question of how these ignored files from external libs find their way into the Assets folder of UltraStar Play.

achimmihca:
The Travis build failed. CI also needs to fetch any used libs before a build.

achimmihca:
I created a shell script to download the needed files of LeanTween directly from GitHub via a sparse-checkout. After the download, the script moves the files to the correct position such that they can be used in our project.

The idea is to run the `download-dependencies.sh` script to have all dependencies downloaded from GitHub. Thus, the Travis build agent must run this script before it does its test. Otherwise the needed libs will be missing.

Windows users can also execute this script, for example by using git-bash.

I think this approach should be easy enough for other devs. It also enables us to use the latest version of the libs from GitHub (or some released version by pulling a tag instead of master). With this script, we do not need to maintain a dedicated repo for the UltraStar dependencies.

PR #60, opened 2019-11-21
improvements

achimmihca:
Several improvements on the dev side via Editor scripts.
There are also some minor fixes regarding layout.

- EditorWindow that shows a list of all scenes for quick scene changes.

- MenuItem to expand the selection in the hierarchy view recursively.

- a C# class is generated with the I18N keys by a MenuItem. Using this class, one does not need to handle raw strings and has autocompletion when using I18N from code.
Furthermore, I added a dropdown with the available keys when working on I18NText components in the inspector.

- to make it easier to position RectTransform such that they will resize for all resolutions, I added MenuItems for "move anchors to corners" and "move corners to anchors".
Using these MenuItems, I fixed several layout issues relatively easy.
Before, there were scripts called AnchorsFollowCorners and CornersFollowAnchors. However, adding these to GameObjects was tedious and they were messing with each other and were not as easy to use. A shortcut on the MenuItem is a much better approach.
Thus, I removed the old two scripts, along with all of their instances on GameObjects. This includes the scenes as well as the Prefabs. As a result, there are changes in a lot of files (namely the scenes and prefabs)

- new MenuItems to select all GameObjects in the scene (or recursively in the current selection) that have a missing script.
These MenuItems made removing the instances of AnchorFollowCorners and CornersFollowAnchors in scenes and prefabs relatively easy.


basisbit:
I am quite impressed by the fast progress and also by how well Unity works for such a game

PR #61, opened 2019-11-24
fixes

achimmihca:
A lot of small fixes and stability improvements.
The game is now pretty stable and I can start most songs in my song lib. Therefor, I had to make the song parsing ignore some inconsisties in txt files.

Duet is working as well. If a single player starts a duet song, then all voices are merged and the single player will sing all voices.

I made a lot of stuff serializable and had to add setters to some properties (because fields that have only a getter are not serialized). Unity does not serialize Dictionaries as well. One has to make a non-generic SerializableDictionary for this to work.
Combined, these things fixed several issues that I had with hotswapping. So, hotswap should work now in all scenes.

Anyway, I implemented most of what I had in mind when I started contributing.
If the pitch detection would work well, then UltraStar Play would be very well playable already.
So, I will try again to find a better pitch detection lib.


basisbit:
@achimmihca could you please take a look at the new codacy issues? There are a few which look like they might be easy to fix.

achimmihca:
Done. What is still left (TODOs and commented code), I don't care about.

Issue #12, opened 2018-06-25
implement microphone input & pitch detection


basisbit:
detect what pitch(es) was "sung" by the player(s) since last time it was checked (polling from some pitch detection service per microphone).

- [x] support multiple input devices
- ~~support multiple channels per device~~ -> #85 
- [x] support getting the best pitches guess for certain time lengths (depending on beats per minute of the current song file)
- [x] detect and properly handle sudden input / output devices being unplugged or added or locked
- [x] check if unity actually detects mics being changed
- [x] support linking mic to player by having the player make noise in front of the desired mic
- ~~auto-detect system total audio latency by playing 3 different pitches and checking for each mic how much latency there is until reception of these tones~~  -> #86 
- [x] audio input -> pitch -> sung note between two beats has to be framerate independent. Thus, don't just rely on the Update() method
- [x] use [Application.RequestUserAuthorization](https://docs.unity3d.com/ScriptReference/Application.RequestUserAuthorization.html) before accessing webcam or microphone

basisbit:
regarding pitch detection:
- vocaluxe uses the solution from performous.
- performous uses a deep recursive FFT algorithm, which might be too slow when implemented in c# - should be tested and measured eventually. See https://www.eetimes.com/document.asp?doc_id=1275415
- performous code is GPLv2+
- autocorrelation seems promising because it is easy to implement and also fast
- more advanced, based on autocorrel.: https://asa.scitation.org/doi/10.1121/1.1458024 or http://www.cs.otago.ac.nz/tartini/papers/A_Smarter_Way_to_Find_Pitch.pdf
- windowing is important for the pitch detection quality but rather difficult to do. Maybe just using the songs beats and calculating pitch for first third, second third and last third, then adding them together but using a 1:3:1 or 1:2:1 weight ratio and using trigonometric functions / modulo % (360°/11) would result in a usable "fair" pitch result for such a karaoke game. This static window size might be a problem for very low pitch singers, but I doubt a human can sing that low,

daniel-j:
I find aubio's YIN implementation good for pitch detection. C library.

achimmihca:
The Unity API seems to slow for real-time (i.e. around 33ms at 30 FPS) Microphone Input.

The bottleneck seems to be
[...]

PR #64, opened 2019-11-25
pitch detection with aubio

achimmihca:
### What does this PR do?

Replace the old C# pitch detection lib with aubio and a C# binding of its API, called Aubio.NET.

Note that I added the needed potion of Aubio.NET into the repo. As alternative, one could create a download script again for this lib. However, the project seems more or less abandoned.
If we do not want to include the Aubio.NET files directly, then I would suggest to make a fork of the original repo, merge its develop branch in the master (because the master does not contain the requried pitch detection API), remove the stuff that we do not need, and then pull this repo into UltraStar Play as dependency.

Note further that I added the aubio.dll that I found from a windows-64bit build of aubio-0.4.6.

Aubio: https://github.com/aubio/aubio
Aubio.NET: https://github.com/aybe/aubio.net/tree/develop

### Motivation

The old pitch detection was not giving good results. Aubio implementes several pitch detection algorithms. I use the Yin algorithm in this pull request, which gives very good results already.

### Additional Notes
I was really lucky to find Aubio.NET! Otherwise, I would not have known how to speak with aubio, which is a C library.

The game is actually playable now.
Maybe one could fine tune the pitch detection algorithm parameters, but it is working very well already.


basisbit:
I reviewed the changes. The code looks good.
Regarding codacy, if you don't want to put curly braces around nested statements in one-liner 'if' blocks, the codacy rule can be disabled - if that was not on purpose, that would be an easy fix.
Regarding travis-ci: build works fine but there apparently are two failing editmode tests which test the song parsing code and expect exceptions - these two tests now can be just deleted.

At last, I am not entirely sure about the legal implications of including a GPLv3 licensed library and linking to it at runtime. We should discuss this in #12 and merge once there is some agreement or the issue is solved.

achimmihca:
Glad to hear that automatic build is working again.
90% of the code issues are in the Aubio.NET implementation, which is not an option anymore anyway.

You can ignore this pull request. I will create a new one without aubio and its dependencies.
I will only include some refactoring to swap the pitch detection algorithm more easily. And I will remove the failing tests.

PR #66, opened 2019-11-28
pitch detection refactoring

achimmihca:
As promised, here comes a refactoring of the MicrophonePitchTracker that makes it easier to change the pitch detection algorithm.
To change the pitch detection algorithm, one has to provide a different implementation of IAudioSamplesAnalyzer.

The currently used pitch detection algorithm (which uses an external C# library called PitchTracker) has thus been separated from the MicrophonePitchTracker into the class CSharpPitchTrackerLibraryAudioSamplesAnalyzer.

Also, I moved the PitchTracker library into the Plugins folder, such that it is not re-compiled every time.


achimmihca:
The codacy issues are intended this time.

achimmihca:
I also removed the failing tests. So I hope travis build will be successful.

basisbit:
Travis-ci tests passed, but build failed because "Error building player because build target was unsupported". That error message is wrong, but instead the project build settings seem to be invalid for that unity version. I guess I'll have to look at the docker image again.... 😅

PR #69, opened 2019-12-01
Replace NAudio usage with native Unity functions

Tom-Julux:
### What does this PR do?

It replace the usage of the NAudio library with native Unity functionality, so the game runs on non Windows systems ie. Linux.

### Motivation

NAudio uses Windows exclusive APIs, so moving away from it at some point is inevitable.

### Additional Notes

The NAudio library is still found under `Assets/NAudio`, but could be removed without breaking anything.

### DIFF
--- a/UltraStar Play/Assets/Scenes/Sing/SingScene.unity
+++ b/UltraStar Play/Assets/Scenes/Sing/SingScene.unity
@@ -588,6 +588,7 @@ RectTransform:
   - {fileID: 163774570}
   - {fileID: 1914020555}
   - {fileID: 1219814412}
+  - {fileID: 1098468896}
   m_Father: {fileID: 0}
   m_RootOrder: 1
   m_LocalEulerAnglesHint: {x: 0, y: 0, z: 0}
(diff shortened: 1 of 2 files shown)


basisbit:
afaik, NAudio or anything similar is required to play mp3 audio files. Also, according to (this)[https://forum.unity.com/threads/importing-audio-files-at-runtime.140088/] using WWW for audio playback is too slow for many platforms. (I guess it loads the whole audio file, and only then starts playback?)

achimmihca:
The PR looks good.
The best way for handling audio is definitely the Unity API because it gives us all the cross-platform support of Unity. 

Regarding MP3: We only have to decode the format somehow. The decoded bytes can be set in an AudioClip, such that the Unity API can be used to play it afterwards.

Maybe we can decode MP3 to WAV with the code of NAudio. There is a good chance that that this portion of NAudio is C# only.
I would start with [Mp3FileReader](https://github.com/naudio/NAudio/blob/master/NAudio/Wave/WaveStreams/Mp3FileReader.cs) and add further code that is needed until it works.

As alternative there is also [Mp3Sharp](https://github.com/ZaneDubya/MP3Sharp).

Futher links for the interested reader:
Loading WAV in Unity
https://github.com/deadlyfingers/UnityWav
https://assetstore.unity.com/packages/tools/audio/open-wav-parser-90832

Convert MP3 to WAV
https://gamedev.stackexchange.com/questions/114885/how-do-i-play-mp3-files-in-unity-standalone
https://stackoverflow.com/questions/11446096/converting-mp3-data-to-wav-data-c-sharp

PR #70, opened 2019-12-02
basic CAMDiff based pitch detection and threshhold loudness checking

basisbit:
### What does this PR do?
Implements fast and good microphone pitch detection.

### Part of Issue(s)
#12 

### Additional Notes
CAMDiff is licensed under public domain. UltraStar Deluxe uses a similar algorithm for pitch detection.


UltraStar Play/Assets/Common/Audio/Recording/CamdAudioSamplesAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
+using System;
+using System.Collections.Generic;
+using UniRx;
+using UnityEngine;
+
+public class CamdAudioSamplesAnalyzer : IAudioSamplesAnalyzer
+{
+    /** There are 49 halftones in the hearable audio spectrum (C2 to C6 (1046.5023 Hz)). */
```
achimmihca: This statement confuses me.
A normal piano keyboard has 88 keys and each give a different sound (and I can hear that sound).
Thus, I expect there are more that 49 halftones, right?

achimmihca: Maybe there are 49 halftones a human can sing, i.e., 49 halftones can be created using a human voice?

UltraStar Play/Assets/Common/Audio/Recording/CamdAudioSamplesAnalyzer.cs
```diff
@@ -0,0 +1,171 @@
...
+    public void Disable()
+    {
+        isEnabled = false;
+    }
+
+    public void ProcessAudioSamples(float[] audioSamplesBuffer, int samplesSinceLastFrame)
+    {
+        if (!isEnabled || samplesSinceLastFrame < MinSampleLength)
+        {
+            return;
```
achimmihca: The samples of this frame are discarded here. I think it would be better to buffer them until the buffer has enough samples to be analyzed.

Otherwise (in theory), if the framerate is so high that the samplesSinceLastFrame are always below the MinSampleLength, then it will never be analyzed.

At the moment the framerate is aimed at 30fps, because videos rarely have a higher frame rate. But this could change in the future.

basisbit: The leftover samples will not be usable for improving pitch detection, and thus can be discarded. Storing these samples in a separate array introduces cpu overhead for partially copying array content, for locking / threaded protected access to some PreviousUnusedSamples array, and either for adding clipping detection or for twice running the camd algorithm (for the "old"/stored part and the new samples).
I'd suggest just discarding the additional samples here, because the quality result will already be good enough here anyways.

What definitely should be done (not part of this pr) is to decouple pitch note averaging from screen frames and instead add all detected pitch values of one usdx file pitch + averaging them, then show that as sung note for that beat duration and grade that averaged pitch comparing to the "perfect" note as mentioned in the song ultrastar txt file. This will greatly improve perceived pitch detection / grading quality especially in noisy environments and increase the amounts of fun 😛 . This is already mentioned as TODO in #12 and (in my opinion) should be done outside of the CamdAudioSamplesAnalyzer.cs class.

basisbit:
@achimmihca I'd like to add amplification and also use the amplification + noise suppression settings - what would be the easiest / proper way to get these settings values?

UltraStar Play/Assets/Common/Audio/Recording/CamdAudioSamplesAnalyzer.cs
basisbit: ah yes, 49 halftones of range (C2 - C6) for typical human singing according to https://en.wikipedia.org/wiki/Vocal_range :

- Soprano: the highest female voice, being able to sing C4 (middle C) to C6 (high C), and possibly higher.
- Mezzo-soprano: a female voice between A3 (A below middle C) and A5 (2nd A above middle C).
- Contralto: the lowest female voice, F3 (F below middle C) to E5. Rare contraltos possess a range similar to the tenor.
- Tenor: the highest male voice, B2 (2nd B below middle C) to A4 (A above Middle C), and possibly higher.
- Baritone: a male voice, G2 (two Gs below middle C) to F4 (F above middle C).
- Bass: the lowest male voice, E2 (two Es below middle C) to E4 (the E above middle C).

Will fix the code comment in the next commit.

achimmihca:
> what would be the easiest / proper way to get these settings values for the current microphone?

These values are stored in the MicProfile.
The PlayerNoteRecorder is initialized with a MicProfile (Init method). At the moment, only the device name is given to the MicrophonePitchTracker. However, the other values (or the complete MicProfile) could be given to the MicrophonePitchTracker as well.

PR #71, opened 2019-12-03
SingScene: fix issues with wrong calculation of song durations / positions

achimmihca:
### What does this PR do?

Fix some issues.
Conversion of duration-in-millis to duration-in-seconds was using multiply instead of division.
Furthermore, the duration of the song is not known directly at the start of the scene. The song is loaded first and afterwards the duration is known. Because of this, the init of the TimeBar had to be moved.


achimmihca:
I noticed that the pitch detection was not sending an event when the singing ended. I added this in your code.

basisbit:
@achimmihca you might want to check this again. Songs now sometimes quit early (for example when the last third of the song was just reached). They sometimes even quit while the gameplay was paused using spacebar key.

basisbit:
@achimmihca it looks like the changes of this PR broke song playback. There now is lots of stuttering in video playback and the audio of the playback is clipping a lot, too. Could you please test the changes on the position calculation again? thanks!

PR #72, opened 2019-12-04
minor microphone input improvements

basisbit:
### Motivation

- Implemented microphone input amplification
- used threshold and amplification from MicProfile
- AudioSampleAnalyzer improvements
- AudioSampleAnalyzer throws new PitchEvent(0) in all cases when there was a samples frame, but no pitch was detected

### DIFF
--- a/UltraStar Play/Assets/Common/Audio/Recording/CamdAudioSamplesAnalyzer.cs
+++ b/UltraStar Play/Assets/Common/Audio/Recording/CamdAudioSamplesAnalyzer.cs
@@ -5,7 +5,7 @@ using UnityEngine;
 
 public class CamdAudioSamplesAnalyzer : IAudioSamplesAnalyzer
 {
-    /** There are 49 halftones in the hearable audio spectrum (C2 to C6 (1046.5023 Hz)). */
+    /** There are 49 halftones in the singable audio spectrum (C2 to C6 (1046.5023 Hz)). */
     private const int NumHalftones = 49;
     /** A4 concert pitch of 440 Hz. */
     private const float BaseToneFreq = 440f;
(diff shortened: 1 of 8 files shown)


UltraStar Play/Assets/Common/Audio/Recording/CamdAudioSamplesAnalyzer.cs
```diff
@@ -170,4 +169,13 @@ private void OnPitchDetected(int midiPitch)
         PitchEvent pitchEvent = new PitchEvent(midiNoteMedian);
         pitchEventStream.OnNext(pitchEvent);
     }
+
+    private void OnNoPitchDetected()
+    {
+        // no tone detected.
+        // Subscribers (such as the PlayerNoteRecorder) must know that the singing ended.
+        // Therefore, a midi pitch of 0 is interpreted as "no singing".
+        pitchRecordHistory.Clear();
```
achimmihca: IMO a "no pitch" of 0 should not have priority over other pitch values. Clearing the history makes it brittle again.

Imagine a pitch detection stream of midi note [60, 60, 60, 0, 60]. In this sequence the 0 would be irrelevant. The median is still 60. But when clearing the history, the 0 will always show up.

Instead, I would implement OnNoPitchDetected() as OnPitchDetected(0);

UltraStar Play/Assets/Common/Audio/Recording/MicrophonePitchTracker.cs
```diff
@@ -14,12 +14,12 @@ public class MicrophonePitchTracker : MonoBehaviour
     [ReadOnly]
     public string lastMidiNoteName;
 
-    private string micDevice;
-    public string MicDevice
+    private MicProfile microphone;
+    public MicProfile Microphone
```
achimmihca: I would prefer if the variable is named the same as the type, i.e., `private MicProfile micProfile` and `public MicProfile MicProfile`.

UltraStar Play/Assets/Common/Model/Microphone/MicProfile.cs
```diff
@@ -28,4 +28,19 @@ public MicProfile(string name)
     {
         this.Name = name;
     }
+
+    public int AmplificationMultiplier()
+    {
```
achimmihca: In general, Amplification is an int so mapping only a few values is not a good definition. I mean, the AmplificationSlider could just as well allow all positive natural numers.

The value in USDX was given in decibel, which is why I expect it to be in dB as well in UltraStar Play. Is this true, is the amplification in dB? If so, is there no general formula to get from amplification in dB to the AmplificationMultiplier?

achimmihca: As alternative, I would put the value of the AmplificationMutliplier directly in the Amplification field and remove the dB unit in the UI. Then it is just some factor.

UltraStar Play/Assets/Common/Audio/Recording/CamdAudioSamplesAnalyzer.cs
basisbit: I disagree. The purpose of the pitchRecordHistory is to filter out short-time noise which is being detected as pitches. Writing 0s in there completely breaks this mechanism and does not improve pitch detection.
The purpose of the pitch detection code is to detect if a pitch was sung, and if yes, return that pitch.  Returning a new pitchEvent of 0 is already a hack which eventually should be removed from the CamdAudioSamplesAnalyzer . 
Without this pull request, there were bunch of cases where no pitchEvent was thrown at all, thus no result for the grading was available for some times and the pitch text display in the record settings occasionally didn't update properly.
The pitchRecordHistory.Clear(); was added because after the player stopped singing a word, we know that very likely a new syllable will be sung and thus the assumption that pitch will most likely not change doesn't fit there any more.

UltraStar Play/Assets/Common/Audio/Recording/MicrophonePitchTracker.cs
basisbit: changes this as requested

UltraStar Play/Assets/Common/Model/Microphone/MicProfile.cs
basisbit: replaced the mapping by proper calculation of the "voltage" amplification multiplier from the field amplification value in the UI. The calculated value is now cached locally.

Issue #44, opened 2019-10-06
new libs: dependency injection (Zenject, UniInject, ...)


achimmihca:
I want to make use of some libs for the UltraStar Play (let's call it USDP ;) ) codebase.
This is a request for comments. Anyone agrees / disagrees?

**Zenject**
For dependency injection (aka Inversion of Control)
Asset store: https://assetstore.unity.com/packages/tools/integration/zenject-dependency-injection-ioc-17758
Tutorial on GitHub: https://github.com/modesttree/Zenject

This could make wiring up objects more easy. It could also be used to register the manager instances in a central way and inject them where needed.
I want to try it and see if my expectations are met.

basisbit:
Being a Java developer who uses Spring Boot at work, I love Dependency Injection. Sadly, I don't have any experience with dependency injection in the C# world. With such a project as UltraStar Play (USPlay / USP/?USDP?) we should try to find solutions that will work for a very long time, thus keep amount of dependencies low and only use already established ones which have a large userbase and thus will probably continue to be supported for a long time.

Could you comment this regarding Zenject and UniRx?

achimmihca:
Regarding dependency injection I took a look at Zenject, [adic](https://github.com/intentor/adic/), [RapidIoC](https://github.com/cpgames/RapidIoC).

Zenject is by far the biggest and most used DI lib of these. It can do everything that I wanted. It also checks the validity of the injection, i.e., that nothing is missing.
However, this is pretty heavy-weight lib and I think 90% of it will be unused for USDP. The project seems to have lost the focus on DI.

RapidIoC requires the implementation of an interface IView, which I consider unlucky naming. For me, a view is something in the UI or a specific view on data.
This project does not seem to have any [community](https://www.chillpillgames.com/community.html?p=%2F%3Fembedded%3D1%26embedCookiesBlocked%3D1) behind it.
From the examples that I saw, injection is only done in children GameObject. Thus, this lib dictates the object hierarchy, which I consider a no go.

Adic is a lightweight lib. However, it seems unmaintained and I already had an [issue](https://github.com/intentor/adic/issues/99) with it. Also, the doc says that objects are safe to use (i.e. injection has been done) after the Start method. I consider this too late because it leaves no usable startup method. In contrast, Zenject has objects completely injected after Awake.

---
Anyway, after using Zenject for our code base, I am not convinced that it will be of great help.
[...]

PR #74, opened 2019-12-04
UniInject

achimmihca:
### What does this PR do?

Add my dependency injection tool UniInject, which I discussed in #44

### Closes Issue(s)

#44

### Additional Notes

For the review, I suggest to start with the tests of UniInject to get a feeling of how the Injector works and what it can do.
Besides this, the UniInjectUtils and SceneInjectionManager are probably a good starting point.

I tried to comment the code as good as I can. Let me know, if you have questions.

---
I also considered to move this to a dedicated GitHub project. But I think it is easier to adapt UniInject to the needs of USP when it is in this repo. It can be moved to dedicated repo later.


basisbit:
started a review, but will take at least one more day or so. Anyways, this looks/feels so good, that it should not just stay a piece of UltraStar Play, but instead make it its own project at github, so more Unity based games can use it. In the long term, this will also make contributing to each of the projects easier for others. Separating the projects of course can be done once it matured for a few months/years so we get some experience of what works well and what might be best practices.

achimmihca:
Ups, I just rebased the branch on master and force pushed. I did not change anything else. The branch should be able to merge now, but you have to approve it again.

Lesson learned: next time I will merge the master into my feature branch (instead of rebase).

PR #76, opened 2019-12-06
small changes/fixes for pitch detection algo

paradigmn:
### What does this PR do?
- change the processing order to make the "halftone != -1" condition valid
- removed element wise division by "samplesSinceLastFrame" -> doesn't appear to be necessary for pitch detection

### Additional Notes
It could be worth considering to combine CircularAverageMagnitudeDifference() and CalculateBestFittingHalftone() into one method. This would make the correlation buffer obsolete. Granted the performance gain is not that huge, but in time critical applications moving data unnecessarily should be avoided.

I couldn't test any of the proposed changes. Please check beforehand if anything works as expected if you want to apply them.

### DIFF
--- a/UltraStar Play/Assets/Common/Audio/Recording/CamdAudioSamplesAnalyzer.cs
+++ b/UltraStar Play/Assets/Common/Audio/Recording/CamdAudioSamplesAnalyzer.cs
@@ -149,6 +149,27 @@ public class CamdAudioSamplesAnalyzer : IAudioSamplesAnalyzer
         // return circular average magnitude difference
         return correlation;
     }
+    
+    private void OnPitchDetected(int midiPitch)
+    {
+        lastPitchDetectedFrame = Time.frameCount;
+
+        // Create history of PitchRecord events
+        pitchRecordHistory.Add(midiPitch);
+        while (pitchRecordHistoryLength > 0 && pitchRecordHistory.Count > pitchRecordHistoryLength)
+        {
+            pitchRecordHistory.RemoveAt(0);
+        }
+
+        // Calculate median of recorded midi note values.
+        // This is done to make the pitch detection more stable, but it increases the latency.
+        List<int> sortedPitchRecordHistory = new List<int>(pitchRecordHistory);
+        sortedPitchRecordHistory.Sort((pitchRecord1, pitchRecord2) => pitchRecord1.CompareTo(pitchRecord2));
+        int midiNoteMedian = sortedPitchRecordHistory[sortedPitchRecordHistory.Count / 2];
+
+        PitchEvent pitchEvent = new PitchEvent(midiNoteMedian);
+        pitchEventStream.OnNext(pitchEvent);
+    }
 
     private void OnNoPitchDetected()
     {
--- a/UltraStar Play/Assets/Common/Audio/Recording/CamdAudioSamplesAnalyzer.cs
+++ b/UltraStar Play/Assets/Common/Audio/Recording/CamdAudioSamplesAnalyzer.cs
@@ -89,7 +89,7 @@ public class CamdAudioSamplesAnalyzer : IAudioSamplesAnalyzer
         int halftone = CalculateBestFittingHalftone(correlation);
         if (halftone != -1 && isEnabled)
         {
-            // add C5 (A4 + 3 halftones) as offset to the detected pitch
+            // no idea where the +3 is coming from...
             OnPitchDetected(halftone + BaseToneMidi + 3);
         }
         else
@@ -143,6 +143,8 @@ public class CamdAudioSamplesAnalyzer : IAudioSamplesAnalyzer
                         audioSamplesBuffer[(index + halftoneDelays[halftone]) & (samplesSinceLastFrame - 1)] -
                         audioSamplesBuffer[index]);
             }
+            // normalize values for future application
+            correlation[halftone] = correlation[halftone] / samplesSinceLastFrame;
         }
         // return circular average magnitude difference
         return correlation;


achimmihca:
I just merged another branch into the master. Please, pull and resolve the conflicts.

> removed element wise division by "samplesSinceLastFrame" -> doesn't appear to be necessary for pitch detection

>I couldn't test any of the proposed changes

If you did not test, then how do you know the element wise division is not necessary?

basisbit:
Indeed, currently that devision is not necessary because the output of the camd function is only used to find the pitch that best fits. The devision would only be required if we would use the result to show a spectrum graph for each player, which would be neat, but is not there currently.

UltraStar Play/Assets/Common/Audio/Recording/CamdAudioSamplesAnalyzer.cs
```diff
@@ -85,11 +85,11 @@ public void ProcessAudioSamples(float[] audioSamplesBuffer, int samplesSinceLast
         // get best fitting tone
         double[] correlation = CircularAverageMagnitudeDifference(audioSamplesBuffer, sampleCountToUse);
 
-        // no idea where the +3 is coming from...
-        int halftone = CalculateBestFittingHalftone(correlation) + BaseToneMidi + 3;
+        int halftone = CalculateBestFittingHalftone(correlation);
         if (halftone != -1 && isEnabled)
         {
-            OnPitchDetected(halftone);
+            // add C5 (A4 + 3 halftones) as offset to the detected pitch
```
basisbit: This cide comment is not correct. BaseToneMidi + 3 is being displayed as A4 in the game UI. I just didn't find the code that causes the +3 offset yet. Please adjust this cide comment.

paradigmn: @achimmihca The pitch detection searches for the smallest value in the correlation buffer. Dividing this buffer by a constant doesn't change the minimum.

@basisbit Ah ok, so the reciprocal correlation buffer could then be used as a rough approximation of the signal spectrum. Do you prefer to keep this line for future implementation?

As for the offset: I assumed adding C5 to the detected pitch has the purpose of keeping the midi notation valid. If the pitch detection returns 2 (D0), then adding C5 (36) would result in D5 (38), while an offset of A4 (33) would result in B4 (35). If a detected pitch of 0 (+A4+3) would indeed result in an A4 being displayed at UI level, would imply that a detected 0 is actually an A0 which doesn't align with the algo as far as I can tell.

paradigmn:
@achimmihca The pitch detection searches for the smallest value in the correlation buffer. Dividing this buffer by a constant doesn't change the minimum.

@basisbit Ah ok, so the reciprocal correlation buffer could then be used as a rough approximation of the signal spectrum. Do you prefer to keep this line for future implementation?

As for the offset: I assumed adding C5 to the detected pitch has the purpose of keeping the midi notation valid. If the pitch detection returns 2 (D0), then adding C5 (36) would result in D5 (38), while an offset of A4 (33) would result in B4 (35). If a detected pitch of 0 (+A4+3) would indeed result in an A4 being displayed at UI level, would imply that a detected 0 is actually an A0 which doesn't align with the algo as far as I can tell.

UltraStar Play/Assets/Common/Audio/Recording/CamdAudioSamplesAnalyzer.cs
achimmihca: When I look at this comparison of midi note integer, name and frequency: https://newt.phys.unsw.edu.au/jw/notes.html
then I would say BaseToneMidi = 33 and 33+ 3 = 36 and 36 is C2.
So who is right now? Maybe the game UI is not correct. Or BaseToneMidi is wrong (the link above names 69 as concert pitch A4)

basisbit: @achimmihca indeed. I don't understand why when I was coding the camd implementation, it required the +3 to make the Midi code translate it to the correct tone.
If I set halftone = 33 (without the +3), it now shows A1.

so, @paradigmn could you please update your pull request to remove the +3 calculation and remove that  from the comment?

paradigmn: I am not at home right now. I will try to get it done when I am back.

basisbit: @achimmihca please use https://www.onlinepianist.com/virtual-piano to play some notes and see what pitch is displayed in the UI. 440Hz is A1, which is tone 33. See https://de.wikipedia.org/wiki/Kammerton
The link you mentioned uses a different standard.

achimmihca: For the keys named Y and C on the online-piano that you linked above, the UI shows me 57 as A3 and 69 as A4.

basisbit: Click on "Letter Notes" to have the virtual piano show key letters

achimmihca: Yes, the UI shows the correct reltive pitch (A for an A). Only question is if the octave is correct.

When I play a sine wave in Audacity with a perfect frequence of 440 Hz, then this is shown as A4 in the UI. This corresponds to the [Scientific Pitch Notation](https://en.wikipedia.org/wiki/Scientific_pitch_notation) (SPN). This is what I intended.

It is indeed confusing that there are different systems for the notation, I did not know this. Maybe I should add a comment to the code to make it clear. But it is not relevant for the game play anyway.

paradigmn: I am a bit confused at this point. Is the "+3" offset necessary or not?

achimmihca: BTW: The absolute name for a note is calculated in MidiUtils.cs

PR #77, opened 2019-12-07
various fixes

achimmihca:
Sorry for all the changes. This grew bigger and bigger because I always found new bugs and edge cases and needed the previous fixes to work it out.

### What does this PR do?

This one fixes issues with drawing notes, recording notes, giving points, finishing the sing scene and maybe few others.

I added a GridBeatDisplayer which is nice to see where beats are in the song.
Similarly, CurrentBeatGridDisplayer shows where Update is called. This is useful to see where beats are skipped.

Recording notes and giving points is now done per beat.
Player scores do not have a floating point inaccuracy issue anymore. If you sing perfect, you get 10000 points.

I changed the IAudioSamplesAnalyzer a bit: ProcessAudio now returns a PitchEvent, and the PitchEvent can be null. So there is no pitchEventStream given to the IAudioSamplesAnalyzer, and no 0 midi note is emitted as "no pitch".

There was also an issue where perfectly sung notes were not recognized, because I was not comparing the relative pitches of notes.

I also noticed that PositionInSongInMillis was changing during the same frame (i.e. different Update calls in the same frame). This makes comparing these values difficult and things can get weird. So now I calculate this value only once per frame.

### DIFF
--- a/UltraStar Play/Assets/Scenes/Sing/PlayerController/PlayerNoteRecorder.cs
+++ b/UltraStar Play/Assets/Scenes/Sing/PlayerController/PlayerNoteRecorder.cs
@@ -1,7 +1,5 @@
 ﻿using System;
-using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
 using UnityEngine;
 using UniRx;
 
@@ -16,14 +14,12 @@ public class PlayerNoteRecorder : MonoBehaviour, IOnHotSwapFinishedListener
 
     private int RoundingDistance { get; set; }
 
-    private double lastPitchDetectedBeat;
-
     private RecordedNote lastRecordedNote;
+    private RecordedNote lastEndedNote;
+    private int nextNoteStartBeat;
 
     private PlayerProfile playerProfile;
     private MicProfile micProfile;
-
-    private IDisposable pitchEventStreamDisposable;
 
     private MicrophonePitchTracker MicrophonePitchTracker
     {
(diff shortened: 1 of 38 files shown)


UltraStar Play/Assets/Scenes/Sing/PlayerController/PlayerNoteRecorder.cs
```diff
@@ -90,125 +87,159 @@ public List<RecordedNote> GetRecordedNotes(Sentence sentence)
...
             }
-            else
+
+            // The lastRecordedNote could be ended above, so the following null check is not redundant.
+            if (lastRecordedNote == null)
             {
-                // Start singing
-                lastRecordedNote = new RecordedNote(midiNote, lastPitchDetectedBeat, currentBeat);
-                // Debug.Log("Start new note (started singing)");
+                if (currentBeat >= nextNoteStartBeat)
```
basisbit: please integrate this into the outer if

UltraStar Play/Assets/Scenes/Sing/PlayerController/PlayerNoteRecorder.cs
```diff
@@ -90,125 +87,159 @@ public List<RecordedNote> GetRecordedNotes(Sentence sentence)
...
+        }
+        else
+        {
+            playerController.OnRecordedNoteContinued(lastRecordedNote);
         }
 
-        // Only accept recorded notes where a note is expected in the song
-        Note noteAtCurrentBeat = GetNoteAtBeat(currentSentence, currentBeat);
-        if (noteAtCurrentBeat == null)
+        if (targetNoteIsDone)
```
basisbit: please integrate this into the same if check a few lines above

achimmihca: done

PR #80, opened 2019-12-10
Anst/audio wave form visualizer

achimmihca:
### What does this PR do?

Improve the visualization of the samples array from microphone. Can also be used to visualize AudioClip of songs.

I also added Plugins/NLayer/ to gitignore.

There are some changes in scripts for the song editor. These can be ignored and are WIP.
However, I want to avoid another huge pull request when I am done with everything. This is why I want to merge the AudioWaveFormVisualizer already.

### Closes Issue(s)

None

### Motivation

I plan to use the audio wave form in the song editor for navigation and orientation in the song, similar to what can be done in Audacity.


achimmihca:
You can see the wave form here: https://www.dropbox.com/s/6o3yt014q1nudej/2019-12-10%2022_52_23-UltraStar%20Play%20-%20RecordingOptionsScene%20-%20PC%2C%20Mac%20%26%20Linux%20Standalone%20-%20Unity%202019.png?dl=0

basisbit:
regarding wave form visualizer, we should use a shader for visualizing, and some transfer object like a one dimensional texture/material as data transfer object.
Shader code could look like this for example: https://gist.github.com/basisbit/2241784c753d71f1f71b9b2337538e75

This would move the image calculation from CPU to GPU, which will be especially important for low end devices like Amazon FireTV stick.

achimmihca:
This is a nice idea, however I have doubts that it will work.
I noticed that the float array for the shader has a maximum size of 1023 (you get a warning if it is bigger). Further, the array size must be given at compile time.
Even the 1 second of the mic samples are already 44100 samples. So calculating the min and max values of this array cannot be done so easily on the GPU.
I want to create a waveform of an entire song, which is even more samples of which the size is not known at compile time.

So, I would merge the new waveform visualization for now. Doing it on the GPU via shader can be a research topic for another ticket.

As alternative, we can keep the simple FloatArrayVisualizer for now and I will use my AudioWaveFormVisualizer implementation only for the SongEditor.

PR #81, opened 2019-12-15
corrected pitch detection

achimmihca:
### What does this PR do?

Fix pitch detection parameters, namely frequencies and lowest singable note.

I found where the mysterious + 3 was coming from. The midi note of C2 (lowest singable voice) is 36 not 33.
Furthermore, there was an issue in the calculation of the frequencies. The octave offset was not correct.

You can find a good comparison of midi note, piano key, frequency, and note name (english and german) here: https://www.inspiredacoustics.com/en/MIDI_note_numbers_and_center_frequencies

I tested my changes with my guitar tuner and 440 Hz sine wave generated in Audacity. The pitch detection is correct now.

### DIFF
--- a/UltraStar Play/Assets/Common/Audio/MidiUtils.cs
+++ b/UltraStar Play/Assets/Common/Audio/MidiUtils.cs
@@ -1,9 +1,15 @@
 using System;
-using System.Collections;
-using System.Collections.Generic;
 
 public static class MidiUtils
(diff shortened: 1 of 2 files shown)


achimmihca:
BTW: I moved the min and max midi notes to MidiUtils, because I need these values also for the Song Editor

PR #101, opened 2019-12-27
Update CONTRIBUTING.md

achimmihca:
Added section where to find and execute tests.
Added note that common scene objects can be injected via UniInject.
Moved I18N stuff to own wiki page.
Added tips for using Awake, Start, OnEnable, etc.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,6 +103,11 @@ As a result, assets that are used in only one scene are placed in the folder of 
 
 ## Project Insights
 
+### Tests
+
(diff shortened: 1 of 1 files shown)


achimmihca:
I find it a bit weird that I cannot change CONTRIBUTING.md without a pull request and everything.
It makes updating documentation unnecessarily complex.

Maybe we should make more use of wiki pages and only link to these in CONTRIBUTING.md?

basisbit:
usually, updating the contributing.md should only be necessary with code changes (if possible) in the same pull request. The Wiki should be used for user-facing documentation.

The contributing.md should (ideally) not be changed too often and not include all of the developer-facing documentation, but instead only be a basic guide for developers with basic introductions to the most basic code areas and samples.

PR #102, opened 2019-12-28
Anst/context menu

achimmihca:
### What does this PR do?

Added scripts to create a context menu.
Use this for a first, simple context menu in SongEditor.

Also: simplify OnPointerClicked methods. Some checks there were not necessary (I was just doing stuff the wrong way).

### Motivation

Instead of an overloaded toolbar, I want to use context menus for various actions in the song editor. For example, "split note" should be available as context menu item on notes.


achimmihca:
@basisbit I always request reviews from you at the moment. Should I also request reviews from others? I can imagine that it can be a lot of work.

I don't know anyone in the list of reviewers.

PR #108, opened 2020-01-14
Anst/yin pitch detection

achimmihca:
### What does this PR do?

Added YinAudioSamplesAnalyzer as alternative to the simple CamdAudioSamplesAnalyzer.
I also added some small tests.
I only added the YinAudioSamplesAnalyzer without using it (except for the test). It has worse performance than the simple CamdAudioSamplesAnalyzer. But it might give better results.

@basisbit could you please test this with your ingenious comparison? I am curious if this gives more points than the Camd version.

Please merge #110 first.

### Motivation

Improve pitch detection for alpha release.

### Additional Notes

I am no expert in audio processing and give no guarantee that the YIN impl is correct. I read the paper and various existing implementations of it to create my own version.

### DIFF
--- a/UltraStar Play/Assets/Common/Audio/MidiUtils.cs
+++ b/UltraStar Play/Assets/Common/Audio/MidiUtils.cs
@@ -12,6 +12,11 @@ public static class MidiUtils
     public const int MidiNoteConcertPitch = 69;
     public const int MidiNoteConcertPitchFrequency = 440;
 
+    public static readonly double[] halftoneFrequencies = PrecalculateHalftoneFrequencies();
+
+    public static readonly double MaxHalftoneFrequency = halftoneFrequencies[halftoneFrequencies.Length - 1];
+    public static readonly double MinHalftoneFrequency = halftoneFrequencies[0];
+
     public static int GetOctave(int midiNote)
     {
         // 12: "C0"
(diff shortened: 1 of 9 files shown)


basisbit:
without merging #110, over 10 test passes each, camd resulted in an average of 6847, while yin resulted in an average of 6802. So, not much difference.

basisbit:
Apparently, on my setup (using the audio feedback from speaker to microphone driver) best works when setting the delay to ~ 440ms. The problem now is, that before the grading of the last syllables of a sentence happen, it switches already to the next sentence and apparently does not give points for the last syllables of the previous sentence.
averages for camd: 6915
yin: 6337

achimmihca:
I see, then I would not put more work into YIN as it does not seem to be beneficial for us (or I implemented it wrong). Anyway, it was fun implementing it. Not trivial, but still do-able and I needed lots of stuff that I learned in university :)

> The problem now is, that before the grading of the last syllables of a sentence happen, it switches already to the next sentence

Yes, this is stupid. I will open a new issue for it.

basisbit:
ah wait... we don't really know if the result would be better or not. The pitch detection looked like it was less flaky/jumpy or more stable with yin than compared to camd. It might be worth it to look at this code again once the other mentioned issues are solved and the numbers are actually more comparable.

Issue #84, opened 2019-12-18
Lyrics helper: correct placement seems to be dependent on screen resolution


bohning:
Issue type: Bug report

### Actual behaviour

The lyrics helper seems to be correctly placed for e.g. a game resolution of 800x600:
<img width="1680" alt="Bildschirmfoto 2019-12-18 um 14 16 09" src="https://user-images.githubusercontent.com/1497707/71089374-ffe67a00-21a0-11ea-9f3e-6ad55ad11280.png">

However, it seems to be incorrectly placed for e.g.  a game resolution of 2560x1600:
<img width="1680" alt="Bildschirmfoto 2019-12-18 um 14 11 13" src="https://user-images.githubusercontent.com/1497707/71089169-91a1b780-21a0-11ea-910c-53c5ad5fde71.png">

It is difficult to see this in the screenshots, but in the correct case, the lyrics helper is always exactly at the current syllable and never goes beyond the last syllable of a sentence, while in the incorrect case, it seems to be "stretched" and to start left of the actual lyrics and goes beyond the last syllable of the sentence (as is obvious in the second screenshot).

### Expected behaviour
The lyrics helper be placed correctly independent of game resolution.

### Steps to reproduce
Start UltraStar Play with different resolutions and observe the position of the lyrics helper.

### Details

Provide some additional information:

- UltraStar Play version: latest trunk
- Operating System + version: macOS 10.15.2

achimmihca:
Please check, whether the issue is fixed with the upcoming built of UltraStar Play.

bohning:
@achimmihca Issue solved, tested several different screen resolutions! Thanks a lot!

PR #117, opened 2020-01-19
Anst/84 fix position in lyrics indicator

achimmihca:
### What does this PR do?

Fix the PositionInLyricsIndicator.

For debugging this issue, I created a DebugPointPrefab that can be created and destroyed via the UiManager.

I also fixed and extended the RectTransformHelper, which was a little broken (anchoredPositionOld was not set).

`Database.json` is now in gitignore, because this will always change and is irrelevant for the code.

### Closes Issue(s)

#84


basisbit:
I didn't see the place where the generate DebugPoints function actually is called. Is this only a convenience function to be added by a developer in code if required, or is that line missing from the PR?

achimmihca:
Yes, the DebugPoint functionality is only for debugging. I used it to find where the target indicator position actually is (and noticed that I was confusing different coordinates). Afterwards I removed the usage.

But the functionality will probably be useful to debug other UI position issues in the future, so I left it.

PR #118, opened 2020-01-26
Anst/parsing fixes

achimmihca:
### What does this PR do?

Replacement for #49 

- also catch FormatException (and all others)
- more strict parsing of #P and #DUETSINGERP tags


achimmihca:
Note that the parsing is now limited to #P and #DUETSINGERP tags that have a single digit number (so #P9 is ok but #P10 would not be accepted).
However, at the moment there is still no agreement on the UltraStar format for a general amount of players. So this should not be a problem (you could say it is undefined).

PR #130, opened 2020-02-15
ContentDownloadScene for downloading tar packed song packages

basisbit:
### What does this PR do?

Adds a new ContentDownloadScene which is accessible from GameOptionsScene.
Users can paste a URI to a song package, see how big the size is and start downloading the file.
During download, a download percentage will be shown, and during unpacking of the tar file, a non-static status message is shown until unpacking of the tar package finished. After unpacking the tar, the temporary tar file will be deleted.

### Motivation

It was rather difficult to download and unpack song content into for the game especially on mobile platforms. This addition changes this.

### Additional Notes

As a next improvement, the Songs folder in the persistentAppData path should be added to the songs folder setting and then then content should be scanned, so that it is directly available in the game without restarting the game.


UltraStar Play/Assets/Scenes/Options/ContentOptions/ContentDownloadSceneController.cs
```diff
@@ -0,0 +1,206 @@
...
+
+            if (request.isNetworkError || request.isHttpError)
+            {
+                AddToLog("Error downloading the requested file!");
+            }
+            else
+            {
+                statusLabel.text = "100%";
+                AddToLog($"Download saved to: {targetPath}\r\n{request.error}");
+            }
```
achimmihca: I have never seen Windows line ending \r\n beeing used in code. Why not use only \n?

achimmihca:
SharpZipLib seems nice. But why not use a download-dependencies script like with the other libs?

achimmihca:
Is Plugins/Threads your own implementation or an external library? I don't understand why it is needed. Does Unity provide no multi-threading?

achimmihca:
Is there a default set for the download url? Is there already such an archive hosted on github?

UltraStar Play/Assets/Scenes/Options/ContentOptions/ContentDownloadSceneController.cs
basisbit: changed that.

basisbit:
>"SharpZipLib seems nice. But why not use a download-dependencies script like with the other libs?"

SharpZipLib only recently changed to MIT and there are [still](https://github.com/icsharpcode/SharpZipLib/issues/103) some people who contributed to old code parts but so far did not sign an agreement for the license change. Also, we only need a small part of the lib. Thus, I cut out the classes needed for unpacking tar archives and its dependencies within the lib. That required updating imports, which would be difficult to do using a download-dependencies-script.

>"Is Plugins/Threads your own implementation or an external library? I don't understand why it is needed. Does Unity provide no multi-threading?"

See the license file in that folder, the code is from @AnsisMalins and it is the simplest and easiest solution for managed background tasks and is already done properly. From what I was told, Unity itself does not offer anything for threading. Of course, we could implement a static thread pool again, but we would end up with something like the ThreadPool class which Ansis already coded.

>"Is there a default set for the download url? Is there already such an archive hosted somewhere?"

Afaik, a ultrastar archive song packages based content sharing ecosystem so far is only implemented by a my little pony fan community and a russian indie music creators community. The https://github.com/UltraStar-Deluxe/songs collection could be shared like this. At 36c3 I talked with some people from cultural commons collecting society and also from Jamendo, regarding sharing legal content sharing possibilities.

basisbit:
My hope is that many more communities will start sharing content using song repositories containing tar archive based song packages. How to create a song package: just tar a song folder, upload it somewhere and you are ready. Most archive tools support the tar format - for example 7-Zip.

Issue #86, opened 2019-12-20
Audio latency detection


basisbit:
Issue type: Feature request

### Actual behaviour
The game currently assumes a certain amount of milliseconds as audio output + input device latency. This fixed latency value does not work for many different hardware configurations and different player distances from the speakers (3ms of extra latency per 1 meter).

### Expected behaviour
The game should play three medium loud pitch tones, count the time from playback started to pitch detected, average the three intervals and use that as pitch as expected latency.
This process should be done once for each in the game activated soundcard and use different pitch frequencies to distinguish the targeted soundcards/microphones.

PR #131, opened 2020-02-17
Anst/calibrate mic delay

achimmihca:
### What does this PR do?

- calibrate mic delay by playing sine wave tones and measuring time until the pitch is detected
- for the mic delay, a new UI control named NumberSpinner has been created. The ItemSlider was meant for few values that can be selected from a list. The NumberSpinner can select any number.
- moved sine wave tone files. Before, they were only used for a unit test.

### Closes Issue(s)

#86 

### Additional Notes

- I seem to have a mic delay of about 330 ms.
- the value changes about +- 30ms, which is 1 frame, which I think is ok.
- at first I thought the value would be way to high, but actually singing feels much more accurate now.


basisbit:
please fix the issues reported by codacy. the rest looks good. This is a really nice feature by the way - thank you for implementing this!

achimmihca:
The remaining Codacy issues (re-using 'newValue' parameter, equality check of floating point) are intended

Issue #92, opened 2019-12-22
Mic visualization in recording options should consider amplification and noise suppression


achimmihca:
### Actual behaviour

The mic visualization in the recording options scene does not consider the amplification and noise suppression.

### Expected behaviour

It would be nice, if these settings are actually reflected in the visualization.

PR #132, opened 2020-02-17
RecordingOptions: visualize noise suppression and amplification

achimmihca:
### What does this PR do?

- recording options mic visualization considers noise suppression and amplification
- mini refactoring of generic Limit method in NumberUtils

### Closes Issue(s)

#92

### DIFF
--- a/UltraStar Play/Assets/Scenes/Options/RecordingOptions/AmplificationSlider.cs
+++ b/UltraStar Play/Assets/Scenes/Options/RecordingOptions/AmplificationSlider.cs
@@ -12,7 +12,7 @@ public class AmplificationSlider : TextItemSlider<int>
     protected override void Awake()
     {
         base.Awake();
-        Items = new List<int> { 0, 6, 12, 18 };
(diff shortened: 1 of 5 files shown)


UltraStar Play/Assets/Scenes/Options/RecordingOptions/AmplificationSlider.cs
```diff
@@ -12,7 +12,7 @@ public class AmplificationSlider : TextItemSlider<int>
     protected override void Awake()
     {
         base.Awake();
-        Items = new List<int> { 0, 6, 12, 18 };
+        Items = new List<int> { 0, 3, 6, 9, 12, 15, 18 };
```
basisbit: why did you add more options here? Don't we want to keep it as simple as possible? I don't think the additional steps will improve peoples gameplay because people are not machines and their distance from the microphones varies a bit while playing.

achimmihca: When seeing the impact on the visualization, I just felt that 6 db was too much, but 0 was too little.

PR #134, opened 2020-02-18
Asynchronous song file loading

basisbit:
### What does this PR do?

- Load discovered txt files in background + update SongRoulette whenever more songs were loaded
- do not scan "license.txt" files. They are not song files.
- improved file not found message.

### Motivation

A frozen game for a minute or so after clicking on `Sing` is not so good for a demonstration of the game. Instead, allow the user to browse through whatever amounts of songs are already loaded, use the search feature and so on, while more songs are being added to the song selection view in the background.


achimmihca:
I think SongMetaManager.FindSongMeta also needs to lock. Please check this. I cannot comment on the line, because it is old code, but now when it is concurrently it might be an issue.
Otherwise, cool stuff!

PR #136, opened 2020-02-20
Anst/small fixes

achimmihca:
### What does this PR do?

Improve some stuff that was bothering me:
- Draw Freestyle notes, Rap and RapGolden differently
- sync video with audio when jumping in the song
- prettier note images in the sing scene. I took the images from USDX. Furthermore I use a HSV shader to set the hue of the images instead of the plain RGB color (which always makes the image darker). This looks much much better.
- a bit of refactoring (SongEditor only)
- when skipping to the next lyrics, then give the player a bit more time (only when not in the Unity Editor).


basisbit:
the coloring of the notes images seems to still need a bit of work. Only the sung notes are shown in player color, but not the notes which the player should sing.
image
Also, the images seem to be stretched too far and thus don't really look round any more.

achimmihca:
>  Only the sung notes are shown in player color, but not the notes which the player should sing.

Both kind of notes are colored. However, red hue is the default. I guess in your screenshot Player02 does not have a MicProfile associated and thus no color. This is also why its avatar image is not colored.

If you change the color of player01 to green, you will see the effect.

> Also, the images seem to be stretched too far and thus don't really look round any more

Could be. The 9-Slice tiling of the images can be changed using the Sprite Editor and by changing the image itself.

achimmihca:
I will look into this. Please wait, before merge

achimmihca:
Ok. I changed the 9-slice of the images a bit. If this is not as desired, then you have to alter the image I think.
Furthermore, I added ImageHueHelper script to set the hue of an image more easily via the inspector as well as programatically.

PR #138, opened 2020-02-22
Anst/fix async song scan

achimmihca:
### What does this PR do?

I had issues with the async song scan stuff. This PR fixes this.
- I search for a default SongMeta by title, for example when opening the song editor directly in Unity. Because of the async loading, it did not find the default SongMeta. So I added a method to wait until the song scan is complete.
- I added SynchronizedList, which just wraps a normal List inside lock-statements. I looked for such a thing but did not find it. C# SynchronizedCollection seems to be unavailable in Unity and the concurrency stuff from C# does not provide a list that can be accessed via index and sorted.
    - Anyway, I replaced the manual locking of songMetas in the SongMetaManager with this implementation. So only the scanLock is left in the SongMetaManager
- scanning songs is not triggered implicitly anymore when getting the song metas. This is too confusing with concurrency. Instead, there is an explicit method to scan-if-not-done-yet.
- moved other generic collection util classes in common folder
- fix NPE in PlayerController

Although I had some issues with it, the async loading works well. I tested it in the SongSelectScene. I added a delay via Thread.Sleep after loading each song and you can see how new songs pop-up one after the other. Nice!


basisbit:
Why not use ConcurrentBag instead of creating the ConcurrentList (there are good reasons why C# does not have one, for example it does not allow for any optimizations)?
The Order does not matter, as the sorting should be done by the UI controller anyways.

achimmihca:
ConcurrentBag also does not have a method to remove a specific single object.
Well, it seems SongMetaManager.Remove is not used anywhere. 

So ConcurrentBag it is.

basisbit:
Another thing: The defaultSong still feels like a hack. Could you please remove it and instead create a new test-scene which (for example) only loads the song you need for testing and then calls the LoadScene with the desired payload? This would also allow us to get rid of the  Debug.LogWarning($"No song with title {defaultSongName} was found. Using the first found song instead.");` log message which will should up in log for everyone but you.

achimmihca:
I don't concur. The defaultSong is only used when the scene was started without SceneData, i.e., when started directly inside the Unity Editor. This is only for development for a fast code-test workflow.
In contrast, when opened from the SongSelectScene (which is the case in production), then the SingSceneData is already defined by the SongSelectScene such that no default is loaded and no message is logged.

In general, I would expect every scene to be startable from within the Unity editor to test that scene. Creating a working default SceneData is a requirement for this.

IMO starting a different scene than the one that should be tested does not make sense. It would contradict the purpose and makes the code-test workflow cumbersome.
Example: I changed something in the Ui of scene A and want to test it. With your proposal I have to change to scene A-test then hit play. Then I see that I made a mistake, so I hit stop. Then I have to go back to scene A again to fix it.
Switching from A to A-test and back to A is useless. Why not start scene A when scene A should be tested?

achimmihca:
One could refactor the construction of the default SceneData into a new class. Until now, the idea is that the XxxSceneController is supposed to load the xxxSceneData, and if it is not set, then it takes care of creating a default.

However, a XxxDefaultSceneDataProvider could do this instead. Then the only-for-development code is better separated from the normal controller code. What do you think?

basisbit:
Not sure how to solve this without adding the overhead for development which you mentioned. Let's just keep it how it is right now and worry about that later / in another pr. Personally, I don't think the  code separation is the problem here.

Issue #113, opened 2020-01-15
"coyote time" or similar player advantage to make the game feel better


achimmihca:
Issue type: Development discussion

### Problem or question to be solved

Pitch detection is still not feeling as good as it could be. This is limiting the fun you have when singing.

### Suggested solution/s

One way might be to improve the pitch detection algorithm. However, improving the pitch detection using YIN did not work as I expected. The current approach is pretty fine already.

Instead I suggest to implement something similar to "coyote time", which gives players an advantage (see https://www.youtube.com/watch?v=sLXLlJ7FhJU at 5:20). The goal is to make scoring feel fair and the player feel like a boss.

One could, for example, delay the scoring by N beats. If the current N beats to-be-scored were detected as "wrong", but the previous and the following N beats were measured as "correct", then also handle the current N beats as if they were measured as "correct".

Such an approach is not costly, and could make our pitch detection feel more stable (although in reality it is not). It makes use of the additional knowledge that we have about the expected pitch.

Good players are not irritated by single false detected pitches that destroy their perfect sentence, but it also does no hand-holding (so if you do not hit a single note, then this does not help you).

achimmihca:
I implemented a simpler version of the proposal from this issue.
A joker is earned for correct continued singing. The joker cancels a wrong beat in continued singing.

I earned about 60 of these jokers when singing a song (so 60 beats were counted as correct although they were not). IMO it makes the pitch detection feel more stable. But it is still not trivial to score points.

PR #142, opened 2020-03-07
Anst/improved pitch detection

achimmihca:
### What does this PR do?

Improve pitch detection accuracy and performance. Also a little bit refactoring.

Performance is improved by the following:
- use float array instead of double array
- do not call extenal method Math.Abs. This was unexpectedly a pretty big improvement, probably because no method stack needs to be generated with return value etc.
- upper limit for analyzed samples
    - There was a self-reinforcing negative performance loop: When FPS dropped once for whatever reason, then more samples where given to the pitch detection, which resulted in bad performance, which means more time per frame, which again means more samples to be analyzed.

Accuracy is improved by the following
**1 - Qualitative comparison of pitch candidates:**
- The calculated correlation can be considered as "error" or "distance" to the checked frequency. Thus, the smaller the error, the more likely the checked frequency matches. 0 error is a perfect match. On the other hand, a big error indicates that the frequency does not match.
- This info is now used to compare the probability of detected frequency candidates. If one candidate has a much smaller error than the rest, then it is clear that this is the best candidate. However, when there are multiple candidates with similar error, then it is not as clear which candidate actually matches the real frequency.
- This leads to a definition for a normalized error for a candidate C: nError(C) = error(C) / sum_error_of_all_candidates
- In the implementation, I take the 3 pitches with lowest error as candidates for this frame.
- Example 1: normalized errors (A4: 0.05, G4: 0.4, B4: 0.55) -> first candidate is clearly the best candidate
- Example 2: normalized error (A4: 0.31, G4: 0.39, B4: 0.4) -> it is not as clear which candidate actually matches
- Now, compare the current candidates together with a history of last best candidates. This prefers a well matching candidate over most recent candidates (this is why I think of it is a qualitative comparison).
- Example of current 3 candidates (index 0 to 2) and last 2 best candidates (index 3, 4): (A4: 0.21, G4: 0.22, B4: 0.3, G4: 0.07, G4: 0.2) -> Amongst the 3 current candidates it is not as clear which one matches best. However, the last best candidate has a significatly low error, which means it was a very likely match. Thus, G4 should be chosen in this frame. Note that the old pitch detection would have choosen A4.
- Finally, in the above example you can see that in the last 2 frames, G4 has been detected as pitch. This is another information that can be used to improve the pitch detection. In the implemenation this is considered via a bias towards recently detected pitches (halftoneContinuationBias). The normalized error is reduced a little for a current candidate when it has the halftone from a history candidate.

The qualitative comparison improves the pitch detection significantly. It replaces the previous median calculation of detected pitches. I needed to recalibrate the mic delay and it dropped from about 300ms to about 200ms. All in all, the delay and pitch detection feels much more similar to USDX now.

**2 - Ignore candidates of notes outside of singable note range**
- The pitch detection tends towards the pitch with the highest frequency when no real signal is given. For example, It always showed C6 when I had noise suppression at 0 and was not making any sounds.
- Thus, I extended the range of checked frequencies with one octave that is outside of the singable note range. If any of these pitches is detected as best candidate for this frame, then use the last detected candidate instead (if possible).

Ignoring notes outside the singable range made the pitch detection less "jumpy".

### Motivation

The YIN pitch detection's performance was not good enough. However, its best part IMO is the "probability" of the estimated pitch. It inspired me to the qualitative comparison of correlation values.

### More

I needed a circular buffer for the pitch candidate history. I took an existing implementation with liberal license from GitHub: https://github.com/joaoportela/CircullarBuffer-CSharp

### Additional Notes

This is the first of multiple PR with improvements. I had to track down various performance issues, but wanted to create different PR for these to make the review easier.

I also have a PR prepared that implements the idea from #113. You might want to wait before re-doing the score comparison with USDX until it is merged as well.

### DIFF
--- a/UltraStar Play/Assets/Common/Audio/Recording/MicrophonePitchTracker.cs
+++ b/UltraStar Play/Assets/Common/Audio/Recording/MicrophonePitchTracker.cs
@@ -9,10 +9,14 @@ public class MicrophonePitchTracker : MonoBehaviour
     // TODO: use 44100Hz (if supported) or fallback to default microphone sample rate
     private const int SampleRateHz = 44100;
 
+    // Longest period of singable notes (C2) requires 674 samples at 44100 Hz sample rate.
+    // Thus, 1024 samples should be sufficient.
+    private const int MaxSampleCountToUse = 1024;
+
     public bool playRecordedAudio;
 
     [Range(0, 1)]
-    public float halftoneContinuationBias = 0.2f;
+    public float halftoneContinuationBias = 0.1f;
 
     [ReadOnly]
     public string lastMidiNoteName;
@@ -91,7 +95,7 @@ public class MicrophonePitchTracker : MonoBehaviour
 
     private IAudioSamplesAnalyzer CreateAudioSamplesAnalyzer()
     {
-        CamdAudioSamplesAnalyzer camdAudioSamplesAnalyzer = new CamdAudioSamplesAnalyzer(SampleRateHz);
+        CamdAudioSamplesAnalyzer camdAudioSamplesAnalyzer = new CamdAudioSamplesAnalyzer(SampleRateHz, MaxSampleCountToUse);
         camdAudioSamplesAnalyzer.HalftoneContinuationBias = halftoneContinuationBias;
         return camdAudioSamplesAnalyzer;
     }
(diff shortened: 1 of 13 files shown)


achimmihca:
BTW: performance improvement because of float instead double and not calling an external method is significant. In a deep profile, the CAMD function dropped from 60ms to 6ms. (Note that during a deep profile everything needs longer because of the profiling overhead)

achimmihca:
Another performance improvement: re-use the arrays instead of instantiating new ones

UltraStar Play/Assets/Common/Audio/Recording/MicrophonePitchTracker.cs
```diff
@@ -9,8 +9,15 @@ public class MicrophonePitchTracker : MonoBehaviour
     // TODO: use 44100Hz (if supported) or fallback to default microphone sample rate
     private const int SampleRateHz = 44100;
 
+    // Longest period of singable notes (C2) requires 674 samples at 44100 Hz sample rate.
+    // Thus, 1024 samples should be sufficient.
+    private const int MaxSampleCountToUse = 1024;
```
basisbit: this should probably not have a maximum, so that setups with higher sample rates still work well. Also, the more samples, the more resilient the pitch detection should be against noise (especially from speakers). The pitch detection code is (or was) so fast, that even in a raspberry pi it was able to to pitch detection for 6 microphones without influencing frame rate. (only in debug-mode it was slow because of missing optimizations)

achimmihca: I am unsure. I would leave it for now, so you can play with this parameter (or remove it) once the other PRs are done.

PR #143, opened 2020-03-07
Anst/improved note recording and drawing

achimmihca:
### What does this PR do?

Part two of improvements.
 - There was an exception because LeanTween did not have more handles for animating. It occured in a song with lots of golden notes, thus lots of golden start to animate. LeanTween.Init is used to double the amount of available handles.
- I am often logging stuff every frame, which is a lot of logs and difficult to read. The LogUtils.LogFrequently method helps me with this. You can accumulate data to be logged and log it once a second (for example).
- There is logic to handle a pitch event also for missed beats. However, drawing these missed beats is not necessary and is now skipped. This was another self-reinforcing performance drain because the more beats were skipped, the more needed to be drawn to catch up.
- An updated UiRecordedNote is now only position anew. Before, it was destroyed, re-instantiated, and then positioned.
- There were some GetComponentsInChildren calls for finding UiNotes and stuff in UiNotes that were not necessary. Some components (e.g. the Text for the lyrics) are set in the inspector instead now.
- a "joker" system for the note recording implements a simpler version of #113. For continued correct singing of beats, a joker is earned. This joker can be used to cancel a missing beat in continued singing. This also makes the pitch detection feel less "jumpy". In tests, I counted something between 60 to 100 jokers for me, which should be fine.

### Closes Issue(s)

#113

### Additional Notes

There is still a small PR in the queue with performance improvements for the video playback.
And another one for audio and video preview in song select.

### DIFF
--- a/UltraStar Play/Assets/Scenes/Sing/PlayerController/RecordedNote.cs
+++ b/UltraStar Play/Assets/Scenes/Sing/PlayerController/RecordedNote.cs
@@ -14,14 +14,6 @@ public class RecordedNote
     public Sentence Sentence { get; set; }
     public Note TargetNote { get; set; }
 
-    public bool IsTargetNoteHit
-    {
-        get
-        {
-            return TargetNote != null && TargetNote.MidiNote == RoundedMidiNote;
-        }
-    }
-
     public RecordedNote(int midiNote, double startBeat, double endBeat)
     {
         this.RecordedMidiNote = midiNote;
--- a/UltraStar Play/Assets/Scenes/Sing/PlayerController/RecordedNote.cs
+++ b/UltraStar Play/Assets/Scenes/Sing/PlayerController/RecordedNote.cs
@@ -14,6 +14,14 @@ public class RecordedNote
     public Sentence Sentence { get; set; }
     public Note TargetNote { get; set; }
 
+    public bool IsTargetNoteHit
+    {
+        get
+        {
+            return TargetNote != null && TargetNote.MidiNote == RoundedMidiNote;
+        }
+    }
+
     public RecordedNote(int midiNote, double startBeat, double endBeat)
     {
         this.RecordedMidiNote = midiNote;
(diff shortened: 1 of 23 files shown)


UltraStar Play/Assets/Scenes/SongEditor/DefaultSongEditorSceneDataProvider.cs
```diff
@@ -35,10 +35,10 @@ private SongMeta GetDefaultSongMeta()
         // The default song meta is for debugging in the Unity editor.
         // A specific song is searched, so first wait for the scan to complete.
         SongMetaManager.Instance.WaitUntilSongScanFinished();
-        SongMeta defaultSongMeta = SongMetaManager.Instance.FindSongMeta(it => it.Title == defaultSongName);
+        SongMeta defaultSongMeta = SongMetaManager.Instance.FindSongMeta(it => it.Title == defaultSongName.Trim());
         if (defaultSongMeta == null)
         {
-            Debug.LogWarning($"No song with title {defaultSongName} was found. Using the first found song instead.");
+            Debug.LogWarning($"No song with title '{defaultSongName.Trim()}' was found. Using the first found song instead.");
```
basisbit: instead of calling songname.trim everywhere, it should only be called once whereever it is read.

UltraStar Play/Assets/Scenes/Sing/PlayerController/RecordedNote.cs
```diff
@@ -14,6 +14,14 @@ public class RecordedNote
     public Sentence Sentence { get; set; }
     public Note TargetNote { get; set; }
 
+    public bool IsTargetNoteHit
```
basisbit: this code is only used in PlayerNoteRecorder. The RecordedNote class should probably be kept as simple and "dumb" as possible (like a Java bean). I'd suggest moving this function to PlayerNoteRecorder or some Utility/Support class?

PR #144, opened 2020-03-09
Anst/improved video performance

achimmihca:
### What does this PR do?

- The SongVideoPlayer synchronizes the playback position with the audio. In the editor it should directly synchronize when jumping forward to a new playback position. However, in the sing scene it should synchronize smoothly. Therefore a flag forceSyncOnForwardJumpInTheSong (false per default) is introduced to prevent forced sync in the sing scene.
- I think the video texture resolution might be a performance issue. I reduced it from 1080p to 720p and I *think* the playback is smoother now with better frame rate. It might be that Unity is scaling up videos with smaller resolutions which results in performance loss. I don't now about others, but most of my videos don't have HD resolution so 720p is enough.
Anyway, this should be further investigated to fit the texture resolution to the video.
- I also set the video depth buffer to `none` instead of `at least 24bit depth (with stencil)`. As I understand, the depth position of a render texture in 3D space would be encoded in this buffer, which we do not need.

### DIFF
--- a/UltraStar Play/Assets/Common/Scene/CommonSceneObjects.prefab
+++ b/UltraStar Play/Assets/Common/Scene/CommonSceneObjects.prefab
@@ -381,6 +381,8 @@ MonoBehaviour:
   m_OnCullStateChanged:
     m_PersistentCalls:
       m_Calls: []
+    m_TypeName: UnityEngine.UI.MaskableGraphic+CullStateChangedEvent, UnityEngine.UI,
+      Version=1.0.0.0, Culture=neutral, PublicKeyToken=null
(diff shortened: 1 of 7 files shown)


basisbit:
nowadays, all the used hardware should support playing 1080p videos rendered by the graphics card just fine. Rendering the video frames is done outside of the main unity thread, thus this is not a problem even on phones or similar. Please change the resolution back to 1080p until some other smarter solution is found.

achimmihca:
You are right. I tested high resolutions (3 * Full HD) and did not see the frame rate drop. So It seems Full HD video is not the source of frame rate issues.

Anyway, I just found a fix for a small issue that was bothering me. The RenderTexture was not cleared before loading and showing a new video, which is why for a moment the last displayed frame of the previous video was visible. I fixed this by clearing (i.e. reset to black) the RenderTexture.

Issue #89, opened 2019-12-22
Preview of music and video in SongSelectScene


achimmihca:
### Actual behaviour

The SongSelectScene is showing covers and song title. However, neither the video nor the music is played at the moment.

### Expected behaviour

Just like in USDX, the video and music should be started somewhere in the middle of the song when a song is selected.

Note that some songs have tags in the txt file that define "previewstart" and "previewend"

basisbit:
For songs where no #PREVIEWSTART or #PREVIEWEND tag is set, the preview should start at some place where there are actually pitches to sing, instead of a random place where maybe there is only some background music for the next X seconds.

bohning:
There’s actually a #PREVIEWEND tag? I have not once seen it in any file.

basisbit:
The song Preview should start only when the last song selction change was at least 0.5 seconds ago, and not before the background-songs-scanning is finished. It is important that the code related to the song preview does not slow down fast scrolling through the songs or the search or the initial song loading.

achimmihca:
> The song Preview should start [...] not before the background-songs-scanning is finished. It is important that the code related to the song preview does not slow down fast scrolling through the songs or the search or the initial song loading.

Song preview does not wait for all songs to be found. This would be inconsistent. Furthermore it is not necessary because the audio is streamed such that impact on performance is minimal.

PR #145, opened 2020-03-11
Anst/song preview

achimmihca:
### What does this PR do?

Implement audio and video preview in SongSelect Scene

- Therefore, refactored AudioManager to set streamAudio flag to true. This flag will stream ogg files etc. for the AudioClip. Now there is no loading lag anymore.
- New option for preview audio volume

Further improvements for SongSelectScene:
- persist selected players in settings
- toggle selected players via `Tab` key
- fixed fuzzy search for songs by holding the `Alt` key.

Other improvements and changes:
- option in settings to show/hide lyrics on the notes
- I played a little with the new pitch detection parameters. I would love some community feedback. Anyway, it should be fine for a first release.

### Closes Issue(s)

#89


UltraStar Play/Assets/Common/Audio/AudioUtils.cs
```diff
@@ -15,43 +18,41 @@ public static AudioClip GetAudioClip(string path)
...
-                }
-                return DownloadHandlerAudioClip.GetContent(www);
+                Task.Delay(30);
             }
+            return downloadHandler.audioClip;
         }
     }
 
-    private static AudioClip LoadMp3(string path)
+    public static AudioClip LoadMp3(string path)
```
basisbit: why did you make LoadMp3 public? It is not used anywhere else as far as I can see.

achimmihca: I needed multiple attemps to find a working solution for async loading of audio. The Unity documentation about streamAudio is not so clear. Anyway, I refactored back and forth between AudioUtils and AudioManager. Looks like I forgot to change this back to private.

Will do with next PR.

PR #148, opened 2020-03-14
Anst/ui improvements

achimmihca:
### What does this PR do?

Some improvements for the UI. There are a lot of changed files, but mostly because I moved stuff, added images, and touched most of the scenes.
- new avatar images
    - I took the ones from USDX but only their silhouette and put it on the USPlay background. This makes it easy to tint the images. Furthermore because there is no detail anymore in the faces etc. it is easier for players to project their own charater onto the avatar image.
- different approach to draw lines in the sing scene. They are now drawn on a texture instead of creating line prefabs. This is the same approach that is also used in the song editor and has better performance.
- more lines in the sing scene to draw notes. Therefore, I implemented a different approach to map midiNotes to their rows because now there can be multiple lines for the same note.
    - I looked at USDX and found that it shows 10 lines on medium difficulty. USPlay now also draws 10 lines.
- new image for the player score background. I took the one from USDX.
- I set the hue of more images to the mic profile color. This includes the score background image and avatar images
    - to change the hue in the singing results scene, I added the PlayerProfileToMicProfile map to its scene data
- darker lyrics background for better contrast to the lyrics
- no player UI background. It was making the background video / imge ugly. Also, USDX does not have one.
- I changed the background image of the scenes to the "plain" version, i.e., where there is no microphone in the upper right corner. This looks better because (a) the mic in the background was distorted depending on the aspect ration and (b) there is now more place for UI controls and it does not distract anymore.
- I removed some unused images, namely the old avatar images and player numer icons (P1, P2, P3, ...). The player number icons are replaced by a single player number background, which can be colored to different hue, and a Text component above it.


achimmihca:
SingScene:
2020-03-14 11_12_38-UltraStar Play

SingingResultsScene:
2020-03-14 11_12_50-UltraStar Play

achimmihca:
Looking at the screenshot, it seems there are now 11 lines. This can easily be changed by setting lineCount in PlayerUiController.

basisbit:
@achimmihca what are the licenses for those silhouette files? Got any source for these?

achimmihca:
As I said, I took the avatar images from USDX, cut out their silhouette, made this silhouette black, and put it on the USPlay background. The USPlay background is MIT so no problem.

The original avatar files can be found here: https://github.com/UltraStar-Deluxe/USDX/tree/master/game/avatars
So question is what is their license? I would assume license of USDX, which is GPL.
However, you once said that USDX files are ok to be used for USPlay. Maybe you can ask the creators for permission?

I will wait with the merge until I have feedback from you.

achimmihca:
For example, the green silhouette in the screenshot is based on https://github.com/UltraStar-Deluxe/USDX/blob/master/game/avatars/male2.jpg

basisbit:
> took the avatar images from USDX, cut out their silhouette, made this silhouette black, and put it on the USPlay background

Oh okay, I somehow missed that. Great job, looks good!

PR #152, opened 2020-03-15
highscore things

achimmihca:
### What does this PR do?

- ~use SongId (i.e. title and artist) instead of SongHash~
    - The song hash made it difficult to see which entry is for which song when looking at the database.json.
    - Furthermore the association between song and highscore should not change when fixing lyrics or modifying notes in the song editor. Thus hashing the body as song id is not a good solution.
    -  The identifying fields of a song is its **title** and **artist**. Instead of a hash, these fields can directly be concatenated to create an Id. This has the advantage that the id is still readable by humans when looking at the database.json.
    - Assuming an artist has created different versions of the same song (e.g. an acoustic or live version), then this can and should be reflected in the title of the SongMeta (e.g. "The Sweet Song (live)").
- added highscore scene
    - top 5 entries, like in USDX
    - opened after the singing results scene, but only if there are highscores to show
- fix highscore sorting (ascending vs. descending)
    - the top score in the song select scene was showing the lowest highscore, because it was sorted ascending and taken the first. It is now sorted descending.
- refactored displaying highscore in song select (SongSelectHighscoreDisplayer)
- record new highscore when closing sing scene after the last note has completed
    - thus, early exit of the song (e.g. in the middle or beginning) will not record a highscore
    - however, exiting the song during the outro is ok, as long as the last note has completed
- database.json is not pretty printed anymore. The whitespace is a considerable amount in this file and it will be a lot bigger than the settings.json. There are various tools to view and format JSON, e.g., plugins for Notepad++ or Firefox browser (works also with XML if you should ever need this).

### Additional Notes

USPlay now has pretty much all the features that I use in USDX. Yay!


basisbit:
Sorry, but SongHash was used on intention. I can't accept this pull request until we have a discussion about this.

achimmihca:
I undid the SongHash change and opened a discussion

PR #156, opened 2020-03-21
Anst/assembly definition files

achimmihca:
### What does this PR do?

- Split compiled code into multiple assemblies for faster compilation. It uses [asmdef](https://docs.unity3d.com/Manual/ScriptCompilationAssemblyDefinitionFiles.html) files of Unity.
- To measure compilation time I used a new plugin, namely [CompileTimeTracker](https://github.com/DarrenTsung/DTCompileTimeTracker).
    - This is a pretty simple addition that measures and logs compile time. Useful to decide when compilation / code structure needs optimization. Also useful to find out how much better compilation has become via an optimization.
- Because the dependencies are more strict, I had to move some files.
    - Scenes are designed to be independant of other scenes. SceneData has been designed as interface for across-scene communication. Thus, the SceneData had to be moved to the Common assembly.
    - SongVideoPlayer and SongAudioPlayer are used by SingScene and SongEditorScene. These have been moved to Common assembly.
- The button callbacks that have been set from within the inspector have been lost (e.g. a reference to SceneNavigator.LoadScene). Setting such button callbacks is brittle because Unity will not tell when they break. Thus, I replaced all these callbacks with proper scripts that subscribe to OnClickAsObservable.

### Motivation

- Compilation took longer and longer.
    - With the new assembly structure compilation now takes about half the time when making a change for a scene. On my machine, this means it dropped from some 15 seconds to some 7 seconds. This is significant when you do a lot of compilations every hour.
    - Changes in the Common assembly still need some 15 seconds to compile.
- Besides this, the stricter separation of assemblies enforces encapsulation, which in my opinion a a positive thing. For instance, the normal game scenes should not have references to the SongEditorScene, which is now enforced.

### More

Assembly dependencies overview:

plantuml graphic

### Additional Notes

Note how cool it is that the plantuml file from inside the repo can be rendered by a remote server and embedded in any GitHub markdown page, such as documentation or PR. I think it's awesome!


achimmihca:
travis-ci is "offline for maintenance"

Issue #157, opened 2020-03-21
Video of song preview not visible when coming back from sing scene


achimmihca:
### Actual behaviour

Video preview stops working after exiting the sing screen and coming back to the song selection screen.
However, it becomes visible again when resizing the game window.

### Expected behaviour

The video preview should be visible without resizing the game window.

### Steps to reproduce

1. go to song select and scroll to a song with a working video preview
2. start singing the song
3. go back to the song select scene. Now, no preview video is shown anymore, also for other songs.

### Details

- UltraStar Play version: current master
- Confirmed on MacOS and Windows.

### Implementation details
So far I was thinking that it might be some Unity bug and hoping it will be fixed when upgrading to Unity 2020. But it could just as well be wrong handling of video texture from our side. I dont know.

There are other video texture issues that might be related to this bug. For example the pixels of the video area in the song editor are sometimes not updated when changing from song editor to sing scene. When this happens, the background is still visible in the right corner.
This render issue as well goes away when resizing the window so it might be related.

achimmihca:
The shader is wrong. Currently, the shader is Unlit/Texture, which seems to have important fields missing for the UI. 
It is working correctly when using UI/Unlit/Transparent. This also has the advantage that the transparency of the video can be set for the fade-in effect of the video preview.

Will be fixed soon.

Issue #155, opened 2020-03-19
Video of Song Preview is distorted


achimmihca:
### Actual behaviour
The video preview in the song select is distorted and sometimes does not fill the whole rect.

### Expected behaviour
The whole rect should be filled and the aspect ratio of the video should be preserved.

### Steps to reproduce
1. Open song select and scroll to a song with a supported video format
2. wait a little until the preview shows up

### Details
- UltraStar Play version: current master
- Operating System: all

### Implementation details
The VideoPlayer component's aspect ratio is set to [FitOutside](https://docs.unity3d.com/ScriptReference/Video.VideoAspectRatio.html), which should be correct.

However, it seems the size of the render texture influences the aspect ratio. Currently it is set to 720x480 px. When setting it to 720x720 px the distortion seems to be worse. This should be investigated.

achimmihca:
![grafik](https://user-images.githubusercontent.com/14967033/77106652-bc114400-6a1f-11ea-80b5-f7da703ab982.png)

achimmihca:
This is two issues.
First, the width and height of the render texture must be the same such that it has an aspect ratio of 1:1
Second, the song audio player in the scene must preserve the aspect ratio of 1:1. Until now, it was scaling with the UI and thus stretched on wide screens. The wrong aspect ratio of the render texture must have countered this. I guess this is why making the render texture width and height equal was making it worse.

Will be fixed soon.

PR #159, opened 2020-03-21
Fix video issues

achimmihca:
### What does this PR do?

- video materials now use shader "UI/Unlit/Transparent" to fix render issues
- video preview has a transparency fade-in effect. This does not start before the video has been loaded and a frame can be shown
- SongSelect: fixed aspect ratio issues of video preview

### Closes Issue(s)

#155 
#157


bohning:
Thanks for fixing this so quickly, @achimmihca !

PR #160, opened 2020-03-22
Anst/overwrite settings from command line

achimmihca:
### What does this PR do?

As discussed in gitter
- You can pass --settingsPath to point to another settings location.
- Furthermore, you can specify --settingsOverwriteJson, which will overwrite anything that has been loaded from the normal Settings.json. Example: --settingsOverwriteJson {"GraphicSettings":{"useImageAsCursor":false}}.
- comparison of command line arguments is case-insensitive, such that --settingsOverwriteJson, --settingsOverwriteJSON or --SettingsOverWriteJSon would all work
- The ApplicationManager has a new list of strings, which is used in the Unity Editor to simulate command line arguments.

### Motivation

Was requested on Gitter. I already use it to point my "production" USPlay to my full song library instead of the few songs that I use for development.

### DIFF
--- a/UltraStar Play/Assets/Common/ApplicationManager.cs
+++ b/UltraStar Play/Assets/Common/ApplicationManager.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections;
+﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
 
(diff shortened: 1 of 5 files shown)


achimmihca:
I updated the documentation

achimmihca:
I just updated the documentation with a page about "command line parameters"

Issue #150, opened 2020-03-15
Do not preview video types that are not supported


barbeque-squared:
Issue type: Bug report

### Actual behaviour
On the preview screen:
* If a song has a video, but Unity [on the platform it's running] does not support that video type, the cover thumbnail thing becomes a black square
* There are no lines in the log related to videos on the preview screen

On the song screen:
* The background goes black for a split second, before falling back to a static background image (if any; I only tested this with songs that also had background images)
* Relevant log lines for the song screen:
```
[Player01] start (or continue) singing of スマイル at 0 ms.
 
(Filename: ./Runtime/Export/Debug/Debug.bindings.h Line: 35)

Writing database
 
(Filename: ./Runtime/Export/Debug/Debug.bindings.h Line: 35)

Start recording with 'WUP-021-0 Mono'
 
(Filename: ./Runtime/Export/Debug/Debug.bindings.h Line: 35)

Video file '/home/bbq/.ultrastardx/songs/Official/[JP][S2E18] Smile Smile Smile/video.mp4' does not exist. Showing background image instead.
```

### Expected behaviour
* If the preview screen cannot play the video, just fall back to the thumbnail (instead of a black square)
* Optionally: since at this point the game _knows_ it will not play the video, don't try it again on the song screen because of the extra flash.

### Steps to reproduce
Use an unsupported video format. In the example above, this was achieved by running a AVC-encoded video in an MPEG-4 container (aka, pretty much your standard .mp4) **on Linux**.

### Details
Provide some additional information:

- UltraStar Play version: Linux build from (I'm assuming) March 15, 2020 ~11:16
- Operating System + version: Archlinux
- Note that the actual log line is also a bit misleading, since (at least in this case) the video absolutely exists. If non-existing and unsupported formats cannot be distinguished, make the logged message more generic.
- Spaces (or strange characters) in path names are not the issue here, I tested it with something that has zero spaces or strange characters in the path, same result.

barbeque-squared:
Can confirm that this is fixed. Very elegantly done, the user experience is much better this way.

basisbit:
Thank you @barbeque-squared for confirming this!

Issue #149, opened 2020-03-15
Better error handling for unsupported audio types


barbeque-squared:
Issue type: Bug report

### Actual behaviour
* No preview audio
* After starting the song, it goes into the song for a split second before going to the highscore screen

### Expected behaviour
Either:
* Get unsupported audio types to work (this is probably not feasible)
* When the game discovers that it cannot play the audio of the preview, either have an error message somewhere and disable starting the song entirely, or show an error popover when trying to start the song.

### Steps to reproduce
1. Take any song that works
2. Encode its music file to (for example) FLAC
3. Update the .txt file accordingly
4. Play the song in USPlay

### Details
Provide some additional information:

- UltraStar Play version: Linux build from (I'm assuming) March 15, 2020 ~11:16
- Operating System + version: Archlinux
- NB: USDX _does_ actually support FLAC. It's fine if USPlay can't, but regardless the error handling should be improved
- if applicable, add a log file:

Relevant log lines when previewing:
```Loading voices of /home/bbq/Programs/UltraStar Play/build-2020-03-15-linux/StandaloneLinux64/Songs/Sefa - Walking in the Air/notes.txt took 0 ms
 
(Filename: ./Runtime/Export/Debug/Debug.bindings.h Line: 35)

Unable to determine the audio type from the URL (file:///home/bbq/Programs/UltraStar Play/build-2020-03-15-linux/StandaloneLinux64/Songs/Sefa - Walking in the Air/song.flac). Please specify the type.
 
(Filename: ./Modules/Audio/Public/AudioClip.cpp Line: 364)

Could not load not AudioClip from path: /home/bbq/Programs/UltraStar Play/build-2020-03-15-linux/StandaloneLinux64/Songs/Sefa - Walking in the Air/song.flac
 
(Filename: ./Runtime/Export/Debug/Debug.bindings.h Line: 35)```

Note that there already is an error there.

barbeque-squared:
The current behaviour is much better. There is no reason to try starting the song anymore (the subtle error is _very_ nice!), and although the popover that comes up you still try to start it, it's easy to get rid of (left/right still work, so you can go to another song, and then close the popover. if you close the popover on a song with broken audio, it'll just popover again)

bohning:
[...]

PR #164, opened 2020-03-27
Anst/better media format errors

achimmihca:
### What does this PR do?

- Do not preview video types that are not supported
    - as fallback, the background image for a song is shown (also with fade-in animation)
- Better error handling for unsupported audio formats
    - An error dialog is shown when attempting to start a song with unsupported audio file
- Some fixes / improvements when audio files / video files do not exist or have other issues

### Closes Issue(s)

#150, #149

### DIFF
--- a/UltraStar Play/Assets/Common/Audio/SongAudioPlayer/SongAudioPlayer.cs
+++ b/UltraStar Play/Assets/Common/Audio/SongAudioPlayer/SongAudioPlayer.cs
@@ -227,6 +227,7 @@ public class SongAudioPlayer : MonoBehaviour
         }
         else
         {
+            audioPlayer.clip = null;
             DurationOfSongInMillis = 0;
         }
     }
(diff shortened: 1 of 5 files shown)


bohning:
Shouldn’t songs with unsupported audio format be skipped altogether when they are loaded (if that’s possible)? What’s the use to keep them in the library when they can’t be played when selected? I’d prefer getting an error message in a log file about all the songs that cannot be loaded due to unsupported (audio) file formats.

achimmihca:
Yes this is a good suggestion. It has to be implemented on a different level, maybe in the SongMetaManager, which could log and skip unkown file extensions. Please create a new issue so we don't forget. 

Note however that the changes from this PR are still valid even when pre-filtering by file extension. Errors can always occur when trying to load a song. For example because of access rights, data curruption, or someone only renaming an mp3 file (or whatever) to ogg.
Loading all song files at the beginning would be too costly, but it is the only way to tell if loading actually works.

PR #165, opened 2020-03-27
mute / unmute audio via F12 (global keyboard shortcut)

achimmihca:
### What does this PR do?

- add shortcut to mute / unmute audio via F12 (global keyboard shortcut)
    - the mute state is not persisted in the settings. It is intended to temporarily stop audio output.


basisbit:
what is the use case for this feature? Afaik, every OS already provides this feature

achimmihca:
Is there a global keyboard shortcut on Windows that mutes / unmutes the audio? If so, is it active per default? I did not find this feature (via keyboard shortcut) 

I have a keyboard with special media keys but these are not available on some keyboards (although many should have them).

I was looking for an easy way to temporarily mute the audio while in-game in fullscreen, without leaving the game.

Issue #127, opened 2020-02-09
Update notification


achimmihca:
A first release of the project is planned. For future releases, it would be useful to show users a notification about new versions. Of course even cooler would be an auto-update feature, but I guess a notification is easier and will do for the moment.

### Where to show the notification

I would show a pop-up dialog in the MainScene. The warning dialog (see UiManager) could be re-used for this.

### Opt-out for new versions

The user should be able to discard the notification with an option to ignore this particular release or all new releases (like "Remind me next time", "Do not remind me again of this version", "Do not remind me again of this version or any other future version").

The "do not remind me again" feature could be a field in the settings. For example `"IgnoreNewVersions":"2020-01-01,2020-02-01"` or `"IgnoreNewVersions":"all"`.

### Detecting new versions

New versions could be detected by downloading a specific file here on GitHub (using raw.githubusercontent.com). The properties file parser could be re-used for this.

For example, the URL `https://raw.githubusercontent.com/UltraStar-Deluxe/Play/release/version.properties` could point to a file containing current release information.

The version naming of USPlay might change (like "Windows Vista" is newer than "Windows NT"), so I suggest to use the release date to check for newer versions.

Example file:
```
version.name = UltraStar Play 1.2.3.4
version.release_date = 2020-06-02
download_link = github.com/blabla.zip
website_link = github.io/blog/USPlay-coolest-newest-version-get-it-now-get-it-Get-It-GET-IT
```

basisbit:
USDX also has such a VERSION file at https://github.com/UltraStar-Deluxe/USDX/blob/master/VERSION
To check if there is a more recent version available, one could simply check if the content of the version file is not equal to the installed version.

PR #167, opened 2020-04-26
new version notification

achimmihca:
### What does this PR do?

- Implement an update notification as suggested in #127.
    - The file with version and build info is called "version.properties.txt" because it is a properties file but Unity only handles txt file extension as normal TextAsset.
Note that the naming scheme of the release version is not strictly dictated. Stuff like "1.2.3", "2017.04+", "2020-04-10", would all work.

- A build time-stamp (yyMMddHHmm) is automatically set in the version file. The version and build time-stamp are shown in the MainScene. However, the build time-stamp is only shown for debug builds to differentiate builds of the same release in development (including now, where there is no release yet).

    - I assume that a granularity of minutes is sufficient in the time-stamp because a built takes about a minute. I assume that two digits for the year is sufficient, because 80 years is a very long time for such an open-source project.

- I just noticed that I also removed the UI stuff for "Show Preview in Song Select" option from GraphicsOptionsScene. It is not needed anymore because the song preview is working really well by now.

### Closes Issue(s)

#127

### Additional Notes

- We have to check if setting the build time-stamp also works when built by Travis CI. It is implemented using the IPreprocessBuildWithReport interface of Unity.
- When the version file has found its place in the repo, then the URL in the NewVersionChecker has to be updated. At the moment it points to a localhost address.


achimmihca:
2020-04-26 19_30_34-UltraStar Play - MainScene - PC, Mac   Linux Standalone - Unity 2019 2 0f1 Perso

achimmihca:
Only the build of MacOS failed. Is the MacOS build new?

basisbit:
macOS build fails occasionally for unknown reasons for various projects. I just restarted it

basisbit:
Could you please add a VERSION file to this repository here and then change remoteVersionFileUrl to the raw file here in the master branch in this repo?

achimmihca:
I renamed the file to VERSION.txt and moved it to the root Asset folder. The txt extension is needed for Unity to recognize it as a TextAsset. And it must be placed in the Asset folder for this as well.

I also changed the remoteVersoinFileUrl such that it points to the address where the file will be located as soon as this PR has been merged.

PR #177, opened 2020-06-18
Anst/more fine grained results scene

### LINKED ISSUE #168: More fine grained results scene
### Current behaviour

At the moment, the results scene shows: total score, score for golden notes, score for normal notes, score for perfect lines.

### Expected behaviour

This could be extended:

- One could show the beats that have been sung correctly
    - Thereby, one could differentiate "good" beats, which are correct because of rounding and "perfect" beats, which are spot on, even without rounding.
- One could show a graph which illustrates the performance over the duration of the song.
    - I suggest to draw on the x-axis the time in the song and on the y-axis the score **at that time** in the song. Thus, the y-axis is ranging from 0 to 10000. Furthermore, the y-axis is never falling, only rising or holding the score (if no points are gained). Extreme cases are 0 points: results in flat line on the x-axis. 10000 points: results in a line that goes straight from lower-left corner to upper-right corner.
    - One could take the end of a sentence as data points. This makes sense because at the moment the score and sentence rating is also shown to the user after each sentence.
    - Such a graph displays nicely where points where gained and where not

Some Vocaloid games called "Project Diva" have such a graph to show song results.
ss_4

achimmihca:
### What does this PR do?

Implement #168 and some other stuff

- Result scene has a button (and keyboard shortcut) to show graph with detailed statistics
- Three bars show the ratio of perfect, good, and missed notes
- A graph shows the scored points over the duration of the song
    - Therefore, I use some classes from a lib called [Unity-Ui-Extensions](https://bitbucket.org/UnityUIExtensions/unity-ui-extensions/src/master/). The license is [BSD](https://bitbucket.org/UnityUIExtensions/unity-ui-extensions/src/master/LICENSE), which allows for distribution, modification. I guess it should be fine together with MIT as long as the BSD license text is included in the distribution. Thus, I added the Unity UI Extensions license text to the CreditsScene.
        - I use UILineRenderer to draw lines between data points
        - I use BestFitOutline, which is a nice outline effect for Text Components.
        - download-dependencies script has been changed accordingly
    - A tooltip on a data point shows some details about the score and time in the song
        - Therefore, I added some re-usable tooltip classes. However, at the moment the favorites Icon in the SongSelectScene is the only other GameObject that has one.
            - The tooltips use some code that has already been used for the context menu. So I refactored the context menu stuff a little into AbstractPointerSensitivePopup.
- To show these additional stats, the PlayerScoreControllerData has been extended
- The SingingResultsScene now uses directly the PlayerScoreControllerData instead of a dedicated, reduced data structure (PlayerScoreResultData has been removed)

There are also some other minor changes
- improved DownloadContentScene. Fixed the buttons and added a button to cancel the download. An URL is provided there as default, so one can download some songs to get started. Also the logging is better now IMO.
- added SearchMethods.GetComponentInChildrenIncludeInactive for UniInject. Was using this at some point and I guess it can be useful in the future again.
- added the editor script MoveParentWithoutAffectingChildren. Because sometimes you want to resize / move a container RectTransform without affecting its children.
- I moved the field of the contextMenuPrefab to the UiManager. Thus, the ContextMenuManager has been removed because it was only holding this one field.

### Closes Issue(s)

closes #168 

### Additional Notes

Uff, this was WAY more work than I had expected.

Drawing such a thing as a simple 2D line in the uUI is more or less impossible. I really tried for hours to draw the line between data points for the graph using Unity's normal LineRenderer.
In the end I gave up and decided to use an existing solution. UILineRenderer of Unity-UI-Extensions really does its job perfectly, just the tool that I was looking for.


achimmihca:
Screenshots:

2020-06-18 21_17_55-UltraStar Play

2020-06-18 00_14_14-UltraStar Play

basisbit:
There exists Unity GL api for simple drawing to textures / materials. See https://docs.unity3d.com/ScriptReference/GL.html

tools/download-dependencies/download-dependencies.sh
```diff
@@ -4,3 +4,4 @@ sh download-unirx.sh
 sh download-fullserializer.sh
 sh download-nlayer.sh
 sh download-csharpsynthforunity.sh
+sh download-unity-ui-extensions.sh
```
basisbit, line 7: please add a newline here at the end of the bash script

achimmihca:
> There exists Unity GL api for simple drawing to textures / materials.

Using the GL stuff will probably create more issues than it will solve, because it messes with the drawing order. For example, tooltip and the red data point would be drawn under the line like everything else would.
Furthermore, I did not find a way to specify a thickness for the line.
Furthermore, the GL stuff does not seem to work with Canvas.renderMode "Screen Space - Overlay".

Anyhow, you do not seem to like the Unity-UI-Extensions dependency.
I thus took another approach and added methods for pixel-based line drawing to the DynamicallyCreatedImage.
Took algorithms from [here](https://de.wikipedia.org/wiki/Bresenham-Algorithmus#Kompakte_Variante) and [here](https://github.com/miloyip/line/blob/master/line_sdfaabb.c). The code from https://github.com/miloyip/line is public domain.

Using these methods, the line is now drawn on a texture, similar to the lines in the song editor.

Have fun!

basisbit:
ahh, the dependnecy was fine. I was only waiting for the newline to be added...

achimmihca:
Ohh, ok.
But I won't change it back now. Please merge when appropriate.

PR #179, opened 2020-08-28
Anst/song select improvements

achimmihca:
### What does this PR do?

- improve song select scene
    - use "\#\<number\>" in search to jump to song with that index
    - add / remove song to playlist via right click context menu. Thereby, playlists inside the playlist folder are available
    - character quick jump buttons at the bottom side (see screenshot). Will jump in the song list to a song with the corresponding character. This also respects the song order (e.g. by title vs. by artist). 
        - "\#" jumps to numbers, "&" jumps to non-alphanumeric character.
        - <del>the available characters are the one found in (my) Android contacts app. For example, X and Q are not included.</del>
    - song order can be changed directly in the song select scene. It is positioned at the lower-left corner, similarly to the playlist slider on the lower-right corner.
    - order-slider, playlist-slider, character-quick-jump are moved into the viewport when the mouse is over a specific tigger area. For example, moving the mouse to the lower-left corner will make the order-slider show up
        - see RectTransformSlideIntoViewport
    - the number of started / finished of a song is show in the UI below the highscore. Fixed the count of finished (when there are multiple voices / players in the song it now counts 1).
        - it is also possible to order by this criteria. This should make it easier to find your own favorite songs
    - persist playlist and song order

- sing scene
    - using the new RectTransformSlideIntoViewport, the FPS display is now invisible per default and can be shown by positioning the mouse at the top center
    - PitchIndicatorArrow optional in settings (only relevant when using NoteStreamDisplayer)
    - NoteStreamDisplayer: width of lyrics up to start of next note, instead end of this note (i.e. use all available space for lyrics)

- image manager: trigger UnloadUnusedResources once for all cached sprites instead of per sprite

- removed unused Boo.Lang import which caused some issue (see Gitter chat)

### Motivation

I had time while in a train :)

Besides this, I have planned some improvements for the song-editor


achimmihca:
CharacterQuickJump:
2020-08-28 19_35_00-UltraStar Play - SongSelectScene - PC, Mac   Linux Standalone - Unity 2019 2 0f1

OrderSlider:
2020-08-28 19_43_36-UltraStar Play - SongSelectScene - PC, Mac   Linux Standalone - Unity 2019 2 0f1

basisbit:
awesome! Thanks for the pull request and all the work you put into this! I'll try to review it this weekend.
PS: I made you admin in this repo because you do much more work here than me by far.

UltraStar Play/Assets/Common/Graphics/ImageManager.cs
```diff
@@ -84,6 +82,8 @@ private static void RemoveUnusedSpritesFromCache()
                 RemoveCachedSprite(cachedSprite);
             }
         }
+        // Trigger garbage collection of used resources.
+        Resources.UnloadUnusedAssets();
```
basisbit, line 86: why does UnloadUnusedAssets get called here? It should not be necessary to call this at all and according to a paid-for fulltime Unity games dev that I just asked, this should not be called at all. Use GemoObject.Destory() (as is already done below in the code) to have the unity code clean up the object and also free up the GPU memory.

UltraStar Play/Assets/Common/Model/ESongSelectOrder.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
```
basisbit, line 1: all those using includes seem unnecessary / unused for the enum.

UltraStar Play/Assets/Scenes/SongSelect/SongRoulette/SongRouletteController.cs
```diff
@@ -232,6 +232,19 @@ public void SelectSong(SongMeta songMeta)
         Selection.Value = new SongSelection(songMeta, songs.IndexOf(songMeta), songs.Count);
     }
 
+    public void SelectSongByIndex(int index, bool wrapAround = true)
+    {
+        if (!wrapAround
+            && (index < 0 || songs.Count <= index))
+        {
+            // Ignore out-of-range index
+            return;
```
basisbit, line 241: This should probably be logged so that faulty code can be found and fixed?

UltraStar Play/Assets/Common/Graphics/ImageManager.cs
achimmihca: I checked it with the profiler and found that you are right. It is not needed. Texture count is still going down when scrolling songs.

UltraStar Play/Assets/Common/Model/ESongSelectOrder.cs
achimmihca: true

UltraStar Play/Assets/Scenes/SongSelect/SongRoulette/SongRouletteController.cs
achimmihca: IMO wrapAround as parameter should ignore out-of-bounds index. For example you might want the previous index and do (index-1). This could be out-of-range but would not be an error from my point of view. It just does nothing if wrapAround is false.

Works as intended.

basisbit (approved):

achimmihca:
I am sorry, you have to approve again.
I messed up the CharacterQuickJumpBar with a duplicate V.

basisbit (approved):

Issue #100, opened 2019-12-27
Implement user input via Unity's new input system


achimmihca:
### Problem or question to be solved

At the moment only keyboard and mouse input is implemented in UltraStar Play.

### Suggested solution/s

In Unity 2020.1 the new input system for Unity will be available in a stable version.

User input should be implemented using the new input system and documented.
I prepared a wiki page for this: https://github.com/UltraStar-Deluxe/Play/wiki/Handling-User-Input.

Issue #33, opened 2018-11-04
Screen division algorithm


barbeque-squared:
### Do you want to request a *feature* or report a *bug*?

Feature

### What did you expect to see?

In theory, we could support _m_ different Voices across _n_ Players. In the best case, and assuming Players do not get to choose which Voice they get to sing, you basically have to make a bunch of rectangles and then cluster them in a visually pleasing way inside a bigger rectangle (ie, the screen).

For the case of making this as generic as possible yet at least somewhat doable, we will assume that the actual number of screens, _p_, is variable as well, but that all screens are the same width and height. Whether menus/background image/video is repeated on all screens or stretched across is outside the scope of this algorithm (could be just a setting), as is the configuration of the screens itself (3x1, 2x2, etc).

So:
*Input:*
* _p_ screens, each of which is width _Sx_ and height _Sy_
* _m_ Voices (depends on song)
* _n_ Players

*Step 1: assign Players to Voices*
* The first Voice will get _ceil(n/m)_ Players assigned to it.
* All subsequent Voices will also get this number of Players assigned to it, until the number of unassigned Players reaches zero.

*Step 2: distribute the Voices and their Players across the screens*
To make optimal use of screen estate, spreading 6 players across 3 screens for a duet song should be possible. Therefore:
* The optimal number of screens per Voice is _SPV = p/m_ (do not round this number)
* If this number >= 1, each Voice gets _floor(SPV)_ screens. Screen 1 gets _ceil(number of Players for this voice / SPV)_ Players, etc. This is important as we need to repeat the lyrics on the different screens.
* Otherwise, each screen gets _ceil(1/SPV)_ Voices, until there are no more Voices left to assign.
In both cases, empty screens can happen.

*Step 3: compute the optimal distribution per screen*
Note: this algorithm will give each Player an equally sized rectangle to put their Notes in, aka if you're with 3 players across 2 screens, the third Player will not get a bigger notes area. If running the algorithm for every screen, for every voice, for every player is okay, then go ahead and do that, otherwise you only need to compute it once for the first screen, the first of its voices, and the first of its players.

*Step 3.1: decide where to place each Voice and its players*
Essentially, this results in a grid. On a portrait screen, it's probably best to have 4 Voices as a column, 2x2 on square-ish monitors, or even as a row if your screen is mega wide. It's probably a function that depends on the aspect ratio of the screen and the number of Voices it needs to fit.

[...]

PR #180, opened 2020-09-16
Unity 2020.1, LayoutGroups, DynamicallyCreatedImage fix

achimmihca:
### What does this PR do?

- upgrade to Unity 2020.1.5
- implement layout in SingScene using (highly unintuitive 🙄) LayoutGroups
    - therefore, added GridLayoutGroupCellSizer to set the cell size of a GridLayoutGroup based on a given number of columns and rows
    - if only 1 player, then the PlayerUi is positioned at the bottom (like in USDX)
    - layout implemented for up to 6 players
        - also in SingingResultsScene. Same method PlayerUiArea.SetupPlayerUiGrid is used to configure the GridLayoutGroup
- after using the LayoutGroups, I noticed issues with the DynamicallyCreatedImage. However, I think these have been there before.
    - the position of RectTransform is not available immediately. I was not able to fix this (tried LayoutRebuilder.ForceRebuildLayoutImmediate, but did not help).
        - so as a workaround, the DynamicallyCreatedImage now waits until its RectTransform has a proper size before creating the texture. (Note that the idea of the DynamicallyCreatedImage is that the image has the pixel size of the RectTransform. It is a canvas for drawing.)
        - before, the default width and height of 256px was used. This is probably the reason why some lines where blurry and did not have the exact color.
    - I think, the texture was never destroyed / released.
        - this could be a reason why Unity got so damn slow after a while (fixing this was the original motivation why I tried to upgrade to 2020.1. I also updated my nvidia drivers...)
- fix position of PositionInLyricsIndicator (the red bar that moves to the lyrics) at start of song.
    - don't know when this broke.
    - the RectTransform is now positioned relative to the left side of its parent, such that localPosition.x=0 is directly at the left side (where it should start).

### Closes Issue(s)

Is this a solution for #33 (maybe not as perfect)? The layout now is more or less the same as in USDX.

### Motivation

- Upgrade to Unity 2020 is requirement for #100 
- also (maybe even more interesting), there is a new option to disable [Domain Reload](https://docs.unity3d.com/2020.1/Documentation/Manual/DomainReloading.html). This actually works (much better than Unity's hot-swap). On my PC the time to start a scene drops from 20-30 seconds to only 2-5 seconds. Really, it is an order of magnitude faster! So I will work on this with the next PR.

### More

- I will document some guide for using LayoutGroups. IMO this is the poorest (and maybe most limiting) part of the Unity UI.


achimmihca:
travis has to be changed for using the new Unity version

basisbit:
looks like 2019.4.x was the last unity version with standalone linux installer images. Getting 2020.1.x to work on travis-ci might be an issue, but well, that should not stop you from continuing development.

basisbit (approved):
looks okay, although I'd suggest testing the layout changes with different screen resolutions and different aspect ratios to see how it will look like on a 4k 16:10 smartphone or an old 800x600p 4:3 projector.

PR #181, opened 2020-09-18
Anst/no domain reload

achimmihca:
### What does this PR do?

As mentioned in the last PR, a new Unity feature allows to disable [Domain Reload](https://docs.unity3d.com/2020.1/Documentation/Manual/DomainReloading.html). This makes start / stop Play Mode 10x faster.
For this to work, static fields have to be reset explicitly.

This PR adds the needed annotations and reset of static fields.
The PR also changes the downloaded dependency of LeanTween to a custom fork, where the static fields are reset.
The other plugins seem to work as is.

### Motivation

Make Unity development fun again :)
Starting / stopping a scene now takes 2-3 seconds instead of 20-30.

### DIFF
--- a/UltraStar Play/Assets/Common/Audio/AudioManager.cs
+++ b/UltraStar Play/Assets/Common/Audio/AudioManager.cs
@@ -11,6 +11,17 @@ using UnityEngine.Networking;
 // Use this over AudioUtils because AudioUtils does not cache AudioClips.
 public class AudioManager : MonoBehaviour
 {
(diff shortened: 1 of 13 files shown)


basisbit:
@achimmihca how about moving that LeanTween fork into the UltraStar-Deluxe github organisation, so that others can also easily contribute there, even when for example you don't continue the work in a few years?

achimmihca:
Sure can do. So far, I did not want to "spam" the UltraStar-Deluxe with such repos.
My CSharpSynthForUnity fork is also a custom one for USPlay. I will fork these into the UltraStar-Deluxe organisation.

basisbit (approved):

Issue #171, opened 2020-05-13
Ask to close game when hitting Escape key in MainScene


achimmihca:
### Actual behaviour

Hitting Escape in the MainScene does nothing.

### Expected behaviour

Like in USDX, hitting Escape key in MainScene should open a dialog that asks whether the game should be closed.

achimmihca:
Also, hitting Escape in other menu scenes should go back to a suited menu.
Maybe this could be implemented by triggering the back button via Escape. A back button should exist in in all menu scenes already.

PR #185, opened 2020-10-03
#171 close game dialog

michielbos:
### What does this PR do?

- Pressing escape (or Android back button) in the main menu displays a dialog asking to close the game.
- Pressing escape in one of the settings/about scenes makes the game go back to the previous scene (linked to the Back button).
- Dialogs now add a grey fade to the rest of the screen, which blocks mouse interaction with other UI elements.

### Closes Issue(s)

#171 Ask to close game when hitting Escape key in MainScene

### Motivation

Because open source is magic. (just like friendship)

### Additional Notes
There are many changes to the scene files due to Unity format updates. The only important change I did to those files was enabling the 'triggerWtihEscape' checkbox on the Back buttons.

There is a strange bug(?) in Unity's GetKeyDown() function that causes it to be called multiple times if called before and after a scene transition. Because of this, holding down the Escape button causes the game to go back multiple scenes. The only workaround I found was waiting for 10-15 frames before listening to GetKeyDown, but that felt too hacky to implement.

Tested with Ubuntu 18.04 and Android 11. One thing I noticed on Android is that the dialogs are relatively small, similar to other text elements on the UI.

### DIFF
--- a/UltraStar Play/Assets/Common/UI/Dialogs/DialogBackdrop.cs
+++ b/UltraStar Play/Assets/Common/UI/Dialogs/DialogBackdrop.cs
@@ -1,0 +1,30 @@
+﻿using System;
+using UnityEngine;
+using UnityEngine.UI;
+
+public class DialogBackdrop : MonoBehaviour
+{
+    public Image background;
+    public Color targetColor = new Color32(0, 0, 0, 90);
+    public float colorTransitionTime = 1f;
+
+    private float startTime;
+    private Color startColor;
+    
+    public void Start()
+    {
+        startTime = Time.time;
+        startColor = background.color;
+        
+        // Make the background match the screen size, regardless of its parent.
+        RectTransform backgroundTransform = background.GetComponent<RectTransform>();
+        backgroundTransform.position = new Vector2(0, 0);
+        backgroundTransform.anchorMax = new Vector2(Screen.width, Screen.height);
+    }
+
+    private void Update()
+    {
+        float progress = (Time.time - startTime) / colorTransitionTime;
+        background.color = Color.Lerp(startColor, targetColor, progress);
+    }
+}
--- a/UltraStar Play/Assets/Common/UI/Dialogs/DialogBackdrop.cs
+++ b/UltraStar Play/Assets/Common/UI/Dialogs/DialogBackdrop.cs
@@ -10,21 +10,32 @@ public class DialogBackdrop : MonoBehaviour
 
     private float startTime;
     private Color startColor;
+    private Vector2Int lastScreenSize;
     
     public void Start()
     {
         startTime = Time.time;
         startColor = background.color;
-        
-        // Make the background match the screen size, regardless of its parent.
-        RectTransform backgroundTransform = background.GetComponent<RectTransform>();
-        backgroundTransform.position = new Vector2(0, 0);
-        backgroundTransform.anchorMax = new Vector2(Screen.width, Screen.height);
+        FitToScreenSize();
     }
 
     private void Update()
     {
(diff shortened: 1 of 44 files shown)


michielbos:
Codacy says the dialogs list should be made readonly (as is usually preferred), but the project guidelines say that readonly should be avoided. Should I make it readonly or leave it as-is?

basisbit:
please don't bump the unity version up, which forces other devs to update. Within a minor version, the project files will be compatible as long as you don't submit the version file.

Regarding the readonly list, @achimmihca what's your opinion on this?

achimmihca:
I would make it readonly.
There are already numerous readonly fields because of codacy suggestions. 
I find the suggestion reasonable. 

We should update the documentation that readonly is not needed to be avoided

basisbit:
I wonder what all those .lighting files are for. Do you maybe know if they are from the new unity version or are used for the transparent background?

michielbos:
Unity generated them for each scene that I modified and saved, apparently it's a new feature in 2020.1...
From what I could find, they're generated for each scene that has lighting settings that differ from the default settings.

https://forum.unity.com/threads/2020-1-new-feature-lighting-settings-asset.832984/
https://docs.unity3d.com/2020.1/Documentation/Manual/class-LightingSettings.html

UltraStar Play/Assets/Common/UI/Dialogs/DialogBackdrop.cs
```diff
@@ -0,0 +1,30 @@
...
+    
+    public void Start()
+    {
+        startTime = Time.time;
+        startColor = background.color;
+        
+        // Make the background match the screen size, regardless of its parent.
+        RectTransform backgroundTransform = background.GetComponent<RectTransform>();
+        backgroundTransform.position = new Vector2(0, 0);
+        backgroundTransform.anchorMax = new Vector2(Screen.width, Screen.height);
```
achimmihca, line 22: anchorMax is relative to parent, where 1 is all of parent width/height.
Setting it to Screen.width would be 1000 times the parent width, but not really independent of the parent.
Furthermore, the anchor is not automatically applied to the corners, so I wonder why this works (if it does for you).

Probably better:
rectTransform.SetSizeWithCurrentAnchors(RectTransform.Axis.Horizontal, Screen.width);
rectTransform.SetSizeWithCurrentAnchors(RectTransform.Axis.Vertical, Screen.height);

achimmihca:
> Within a minor version, the project files will be compatible as long as you don't submit the version file.

Interesting. @basisbit You have a source for this statement?

---
Regarding Unity 2020.1.7: I took a look at the [release notes](https://unity3d.com/de/unity/whats-new/2020.1.7) and some known issues sound pretty bad:
- Scene Management: Allocated memory is not cleared when loading and unloading scenes
- Window Management: [window layout] cannot load editor layout, stuck in infinite loop of "failed to load window layout"
- Packman: Editor crashes when upgrading/downgrading between 2020.1 and 2020.2

Thus, I would like to stay with 2020.1.5 for now.

UltraStar Play/Assets/Common/UI/Dialogs/DialogBackdrop.cs
michielbos: Oops, I feel kind of embarrassed for not noticing. You are correct, the background was scaled to about 1000 times the dialog (which "works" for filling the screen, as long as you don't look at it in scene mode ^^).

Fixed the issue. Also added a check to reapply the resize if the screen size changes while a dialog is open, because that could otherwise cause future issues with for example mobile device screen orientation changes.

achimmihca (approved):

PR #187, opened 2020-10-09
Anst/song editor lyrics editing improvements

achimmihca:
### What does this PR do?

NO PANIC 😉 . Most file changes are because of (a) refactoring (b) using TextMesh Pro.

This is a major PR that implements proper editing of lyrics in the song-editor.
Besides this, there are some small improvements and fixes.

- Editing lyrics
    - LyricsArea works per voice. Buttons at its side are used to change the voice if there is more than one.
    - LyricsArea now uses a "TextMesh Pro" InputField (TMP_InputField). This provides proper scrolling and selection and ships as a normal Unity package. I think the package was already a dependency, but so far it was not used.
    - editing approach is inspired by [songeasy](https://sites.google.com/view/songeasy/home)
        - workflow is: copy the normal lyrics (from the internet), add some characters to separate notes and sentences, paste this (edit-mode-text) into the LyricsArea, done
    - text on notes is mapped to edit-mode-text or view-mode-text in LyricsArea. From edit-mode-text it is mapped back to the notes.
        - edit mode starts when LyricsArea receives focus. It ends when LyricsArea lost focus, then the text is applied back to the notes
        - edit-mode-text uses three semantic characters: **semicolon** separates notes, **space** separates notes and adds a space to previous note, **newline** separates sentences
    - LyricsArea shows the whitespace characters. This is done by replacing whitespace with visible characters and vice versa, back and forth.
    - Finding note corresponding to the caret position is now done by parsing the edit-mode-text
        - (a) find matching sentence and then (b) find matching note in that sentence
        - this way, position of caret is synchronized with playback position in song. You always see which notes you are editing
    - changes to LyricsArea are immediately visible on notes which makes it pretty intuitive
    - There is a new check that gives a warning when a note does not have lyrics

- text on notes wraps vertically when there is not enough horizontal space. This makes sense because in vertical direction there should only be one note which means that there should be enough space for the lyrics.
    - the font size adapts to the width of the note

Besides this
- context menu action to "fit viewport to sentence"
- issues markers use TooltipHandler to show message. Thus, they don't need to implement IPointerEnterHandler etc. themselves anymore
- cross-hatching image to highlight golden notes in song-editor
    - also, the hatching images now tile seamlessly
- SpaceBetweenNotesAction is applied per voice if no notes are selected
- ISongMetaChangeEvent is now an abstract class called SongMetaChangeEvent
    - this is one reason for many changed files
    - it has a property to opt-out from recording the change in the undo history
        - this was needed for the lyrics change to be (a) immediately visible and (b) only the last state (when LyricsArea lost focus) should be kept in the history

- MicProfile and the RecordingDeviceColorSlider now use Color32 (which is 0-255 byte instead of 0-1 float for r,g,b,a)
    - Motivation: I noticed that loaded colors where not recognized as equal because of floating point inaccuracy
    - Switching to Color32 is thus also needed for ThemeableColor etc.
    - If stuff is invisible to you, then its because you load the old Color values and event 1 compared to 255 is pretty transparent and pretty black. 
        - You have to go to Recording Options and select color again for your devices.

### Motivation

It's awesome having a proper editor for UltraStar song lyrics!

### More

- Will update documentation with the new features when I find time

### Additional Notes

- You might find comments about ColorPalettes. I undid this change. Instead I will refine the theme-mechanism and I18N in a future PR.


UltraStar Play/Assets/Common/UI/UiManager.cs
```diff
@@ -37,6 +38,11 @@ public static UiManager Instance
     [InjectedInInspector]
     public Tooltip tooltipPrefab;
 
+    private Subject<Vector3> mousePositionChangeEventStream = new Subject<Vector3>();
+    public IObservable<Vector3> MousePositionChangeEventStream => mousePositionChangeEventStream;
+
+    public bool DialogOpen => dialogs.Count > 0;
```
basisbit, line 44: when does this get evaluated? only when a new UiManager is created?

basisbit:
@achimmihca could you maybe move all the TMP stuff into its own git repository and add that asset as dependency as usual? Also, what is Text Mesh Pro needed for?

UltraStar Play/Assets/Common/UI/UiManager.cs
achimmihca: The expression is evaluated every time the Property is accessed (=> is like a lambda).
It is equivalent to `public DialogOpen { get { return dialogs.Count > 0; } }`

achimmihca:
> could you maybe move all the TMP stuff into its own git repository and add that asset as dependency as usual?

Will try that

achimmihca:
The remaining Codacy issue is a false positive because uiText.fontSize is a property that is set. I noticed flickering when this is done in every frame.

achimmihca:
@basisbit are you short on time at the moment or did I forget to adjust something?
Any news on this PR?

basisbit (approved):

basisbit:
sorry, was sick in bed the last couple of days.

achimmihca:
No need for apologies. Glad you're back 👍

PR #191, opened 2020-11-01
fix iOS Player Settings

achimmihca:
### What does this PR do?

Fix iOS Player Settings that were mentioned on Gitter.

- added microphoneUsageDescription
- checked "mute other audio sources" to enable Bluetooth microphones

### DIFF
--- a/UltraStar Play/ProjectSettings/ProjectSettings.asset
+++ b/UltraStar Play/ProjectSettings/ProjectSettings.asset
@@ -72,7 +72,7 @@ PlayerSettings:
   macRetinaSupport: 1
   runInBackground: 1
   captureSingleScreen: 0
-  muteOtherAudioSources: 0
+  muteOtherAudioSources: 1
   Prepare IOS For Recording: 0
   Force IOS Speakers When Recording: 0
   deferSystemGesturesMode: 0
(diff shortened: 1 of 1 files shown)


achimmihca:
2020-11-01 16_13_07-Project Settings

basisbit:
This seems to be a "global" setting for all build targets. Can't this setting be toggled by code if iOS is detected?

bohning:
It is *at least* required for macOS, not sure about iOS (but I would assume so).

achimmihca:
> [PlayerSettings.muteOtherAudioSources](https://docs.unity3d.com/ScriptReference/PlayerSettings-muteOtherAudioSources.html):
This setting is shared between iOS and Android platforms.
Starting with Android Marshmallow (6.0), setting this to false mutes the sound of your Unity application during an incoming phone call (while the phone is ringing).

Seems reasonable to set it to false for Android because the App should mute for an incoming phone call.
However, the PlayerSettings are in UnityEditor namespace and not available in-game. So I guess there is no way to set it for a single platform.

Anyway, it is not be a big issue on Android.
IMO bluetooth mic device for iOS has higher priority.

---
microphoneUsageDescription: I don't think we need this per platform, right?

basisbit (approved):

PR #193, opened 2020-11-08
Anst/improvements

achimmihca:
### What does this PR do?

Some small changes to iron out issues that I noticed while playing with friends.

SingScene:
- show lyrics of second voice in duett song
    - Copied LyricsDisplayer to top side and added boolean to differentiate the instances. The instances are disabled and connected with PlayerController as necessary.
- auto hide cursor when it is not moved for 5 seconds
- fix: Notes are gone and Exception when entering duett song second time
    - if only one player for a duett song, then the notes are merged to create a single voice. This voice is destroyed later. The notes where not cloned and thus were missing in the original SongMeta afterwards.

SongSelectScene:
- translated no-songs-found message
- remember and restore player-to-microphone assignment
    - the mapping is stored in a static field in OnDestroy(). It is restored in Start()
- log number of successfully loaded and failed-to-load songs

SongEditor:
- fixed font sizes of controls in options overlay
    - the text was invisible because minSize was 10 instead 2

### DIFF
--- a/UltraStar Play/Assets/Scenes/Sing/SingSceneController.cs
+++ b/UltraStar Play/Assets/Scenes/Sing/SingSceneController.cs
@@ -150,8 +150,7 @@ public class SingSceneController : MonoBehaviour, INeedInjection, IBinder, IOnHo
         // Associate LyricsDisplayer with one of the (duett) players
         if (!PlayerControllers.IsNullOrEmpty())
         {
-            LyricsDisplayer lyricsDisplayer = FindObjectOfType<LyricsDisplayer>();
-            lyricsDisplayer?.Init(PlayerControllers[0]);
+            FindObjectsOfType<LyricsDisplayer>().ForEach(it => InitLyricsDisplayer(it));
         }
 
         //Save information about the song being started into stats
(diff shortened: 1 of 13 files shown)


UltraStar Play/Assets/Scenes/Sing/SingSceneController.cs
```diff
@@ -166,6 +165,33 @@ void Start()
         LayoutRebuilder.ForceRebuildLayoutImmediate(CanvasUtils.FindCanvas().GetComponent<RectTransform>());
     }
 
+    private void InitLyricsDisplayer(LyricsDisplayer lyricsDisplayer)
+    {
+        bool needSecondLyricsDisplayer = SongMeta.GetVoices().Count > 1 && PlayerControllers.Count > 1;
+        if (lyricsDisplayer.isTop)
```
basisbit, line 171: this bool is never changed... Did you maybe forget adding some files to the changelist before committing?

achimmihca: The boolean flag was pretty stupid (it was set in the Inspector). I only wanted to differentiate the instances.
I now added references to the instances in the SingSceneController, which also makes FindObjectsOfType unecessary.

basisbit (approved):

PR #194, opened 2020-11-10
Anst/fixes

achimmihca:
### What does this PR do?

- as mentioned on Gitter, the Icons folder was not comitted
    - "Icons?" was in .gitignore
        - this [SO thread](https://stackoverflow.com/questions/17556250/how-to-ignore-icon-in-git) implies that there is some special Icons folder on MacOS. However, the pattern "Icons[\r]" seems to be the better choice for gitignore
    - I committed the missing folder and the alternative pattern

I also noticed some other issues after cloning a fresh version of the repo that I fixed alongside:
- NPE in EditorCoroutineManager if no CoroutineManager is in the scene (happens on first startup with fresh repo)
- "Assets/Plugins/TextMesh Pro" was always shown by git as changed even though this folder is ignored
    - I removed the .git folder inside of it and the issue is now gone
    - Thus, I also removed other .git folders from downloaded repos for consistency
- when doing UniInject check, it was always complaining that Injector instance is never bound. This happens at runtime in SceneInjectionManager. So I now ignore this type for the check.


basisbit (approved):

bohning:
Icons are back again (on Mac)! Thanks a bunch.

PR #195, opened 2020-11-14
Anst/file system watcher

achimmihca:
### What does this PR do?

- add FileSystemWatcher for translations and theme files
    - fix reloading of theme files by clearing image and audio cache
    - ThemeManager restores last loaded theme on reload
- added missing translations in SongSelectScene and SingScene
- fix WarningDialog not showing custom title

### Motivation

Better workflow for translations and themes: Edit file, view in Unity Editor, repeat


basisbit:
FileSystemWatcher does not work on all Unity targets. I'd suggest changing ThemeFileSystemWatcher so that it only does its job when Application.isEditor is true (or when the current environment is a windows/linux/macos device). Would that be okay for the use case?

achimmihca:
Both file watchers are in the "Editor" special Unity folder and are doing their job only in the Unity Editor.
`using UnityEditor; ` works only in the Unity Editor and will not work in a build. The file watchers are using this to register the `EditorApplication.update` callback.

So, no Application.isEditor is needed.
They are not intended to be part of the game, only for development.

basisbit (approved):

PR #203, opened 2020-11-21
fix NewVersionChecker

achimmihca:
### What does this PR do?

The NewVersionChecker was comparing the numbers of a version string by simply removing all non-digit text.
However, this is not good, because it will consider "1.1.1" smaller than "1.01.01" (because 111 < 10101).

So now
- compare each number separately
- added test for version comparison
- editor tests can now access scene classes via new asmdef file
    - removed test assembly from "normal" editor assembly
    - updated assembly-dependencies documentation file (plantuml)

### DIFF
--- a/UltraStar Play/Assets/Scenes/Main/CompareVersionException.cs
+++ b/UltraStar Play/Assets/Scenes/Main/CompareVersionException.cs
@@ -1,0 +1,8 @@
+﻿using System;
+
+public class CompareVersionException : Exception
+{
+    public CompareVersionException(string message) : base(message)
+    {
+    }
+}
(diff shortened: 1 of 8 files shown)


UltraStar Play/Assets/Scenes/Main/CompareVersionException.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System;
+
+public class CompareVersionException : Exception
+{
+    public CompareVersionException(string message) : base(message)
+    {
```
basisbit, line 6: shouldn't this contain some logging?

achimmihca: nope. its just the exception class. the call base() will set the message in the base-class (like super in java). the code where the exception is catched has to do the logging - and it does.

basisbit (approved):

PR #206, opened 2020-11-25
Show recorded note of all beats and better HSV shader for image

achimmihca:
### What does this PR do?

- Option to show recorded note for all beats, i.e., the beats that do not have a target note
    - I think this helps the player to find the correct pitch.
    - Also (maybe just for me), it makes the pitch detection feel more reliable
- Previously I assumed that a RecordedNote always had a TargetNote. This is not the case anymore.
    - null checks have been added
    - The sentence of the RecordedNote is stored in a new field

- I wanted to draw these recorded notes transparent and without color (0 saturation). Thereby, I realized 
    - (a) the HSVRangeShader did not allow to set the saturation, value, or transparency
    - (b) the shader could be much simpler because all I want to do is apply the color in HSV spectrum. Not shifting colors around.

- So I refined the HSVImageShader to just apply the given color in HSV spectrum.
    - Thus, the ImageHueHelper and ThemeableHue and stuff is not needed anymore and has been removed.
    - I don't know if the MIT text of the original implementation is still needed. However, it does not hurt either.


achimmihca:
analyzed-beats-without-target-note

Hsv-color-picker-and-shader

basisbit:
nice to see that shaders have been used here. That will be crucial for supporting lower-end devices

basisbit (approved):

PR #216, opened 2020-12-22
Serilog for logging

### LINKED ISSUE #207: Logging ?
### Problem or question to be solved
The alpha project names "Logging" as a requirement (https://github.com/UltraStar-Deluxe/Play/projects).
What exactly is desired here?

So far, some logging is done via Unity's Debug.Log methods.
Also, something like this has to be used for logging that should be visible in the Unity Editor's console view

achimmihca:
### What does this PR do?

Add Serilog library for logging.

I researched logging libs for C# and found that NLog and Serilog are modern tools for this task.
I tried both. NLog has a big repo (5 MB only the source folder) with everything in it and I got lots of compiler issues.
In contrast, Serilog has a dedicated repo for every sink (i.e. logging target) such that the repos themselves are pretty small (700 KB total for core and file-sink source). I also had only few compiler issues.

Thus I chose Serilog.

- Added define symbol OS_MUTEX for Serilog compilation
    - This defines which SharedFileSink implementation is used [OsMutex](https://github.com/serilog/serilog-sinks-file/blob/06e457a3d19b4a0cb00717a0bfd5c3be2d2016fe/src/Serilog.Sinks.File/Sinks/File/SharedFileSink.OSMutex.cs) or [AtomicAppend](https://github.com/serilog/serilog-sinks-file/blob/06e457a3d19b4a0cb00717a0bfd5c3be2d2016fe/src/Serilog.Sinks.File/Sinks/File/SharedFileSink.AtomicAppend.cs)
    - I don't really know how the implementations differ, but `shared` parameter in config is false so I guess we do not need this feature anyway.
- Configured Serilog in Log.cs and can be accessed via `Log.Logger`
    - use rolling-file-configuration. Logs are saved to `Application.persistentDataPath + "/Logs"`. Log file changes daily, max 5 files, max 250 MB.
    - For exceptions, the stacke trace is appended (see UnityStackTraceEnricher)
- normal Unity logging is also passed to Serilog (via Application.logMessageReceived)
- Serilog logging is also passed to Unity logging
    - To prevent cycles, a marker is added to the log message (depending on the original source this is `unityLogEventSinkMarker` or `skipUnityLogEventSinkPropertyName`)

This is a initial, basic logging configuration that could be extended to send logs (or only exceptions) to a logging service for analysis, statistics etc. However, I think it is sufficient for our purpose.

### Closes Issue(s)

close #207

### DIFF
--- a/UltraStar Play/Assets/.gitignore
+++ b/UltraStar Play/Assets/.gitignore
@@ -30,3 +30,8 @@ Plugins/TexMesh Pro.meta
 
 Plugins/UniInject/
 Plugins/UniInject.meta
+
+Plugins/Serilog/Serilog/
+Plugins/Serilog/Serilog.meta
+Plugins/Serilog/Serilog.Sinks.File/
+Plugins/Serilog/Serilog.Sinks.File.meta
--- a/UltraStar Play/Assets/Common/ApplicationManager.cs
+++ b/UltraStar Play/Assets/Common/ApplicationManager.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections;
 using System.Collections.Generic;
 using UnityEngine;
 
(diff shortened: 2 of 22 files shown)


basisbit (approved):

basisbit:
@achimmihca could you please add Serilog to /documentation/assembly-dependencies.plantuml ?

Issue #176, opened 2020-06-05
API & App/Web Interface to manage USPlay 


cRaZy-bisCuiT:
Issue type: Feature request

### Features
1. API to access & configure USPlay local network
 - provide an API to control USPlay 

2. App/Web interface to Configure USPlay in runtime
 - Provide a way of configuring USPlay in runtime. We could either provide an App for mobile devices (Android, iOS, etc.) or a responsive (for different screen sizes like Notebook, Tablet, Smartphone) web interface which could be accessible in the local network.

**Features**
Some ideas, what could be implemented/exposed on such a Interface:
- **Queue** -> play those songs after the current one is finished (a searchable library should be exposed to make that possible)
- **change Player Name** for the next songs in advance  -> nice for toplists, so if the mics are rotaded the player name does not stay the same the whole evening for each microphone/colour
- **adjust volume level**
- **activate / deactivate video playback**
- TBA

basisbit:
As far as I know Performous did implement a webapp to do some of those features in a local network. Might be a good start to get inspired by their API.

Anyways, you are mixing UltraStar Play and UltraStar Deluxe a bit there. Could you edit the issue description so that it is clear what you want to control?

cRaZy-bisCuiT:
Sorry for that. Since we're at the USDP repo, Play is meant. Ofc having those features in USDX as well would be nice. I assume the future might be brithter for USDP. Maybe the same code base could be shared on both projects.

basisbit:
*USP or USPlay or UltraStar Play

achimmihca:
From #245, the idea came up that the API and thus Companion App should allow to
- add player profiles
- select songs, players, and their mic for the next round(s)
    - this corresponds to the "queue" feature as suggested above

Vocaluxe also has a similar "Smartphone control" feature.

Baklap4:
Feel free to take a look over at our [api from performous](https://github.com/performous/performous/wiki/Webserver-API)
To be fair it has its problems and isn't the most beautiful api i've written. We have an open issue to refactor it.

achimmihca:
The performous api has some redundant endpoints, e.g., GET "/api/getplaylistTimeout".
The GET http method and get in URL is redundant.


As convention, I suggest to use GET http method for reading stuff,  and POST http method to write stuff

achimmihca:
What should also be considered is some sort of whitelist for devices.

[...]

PR #234, opened 2021-04-02
companion app support

### LINKED ISSUE #230: Companion app for mic input
### Actual behaviour

One of the obstacles for players is the need for a mic, especially when using different devices or mobile.

For example, I bought USB mics just for the game.

However, these mics cannot be used to play the game on a smartphone / tablet.

### Expected behaviour

It would be very nice if a smartphone app could be used to provide mic input for UltraStar Play.
Everyone has a smartphone so people can start singing duets without the need for additional hardware.

Partially, there are existing solutions such as WO Mic. However, these do not work across all platforms and sometimes only for a limited number of devices.

**Mic access**
Accessing the mic is already implemented in UltraStar Play. This part can be copied for the companion app.

**Network discovery**
What is missing is some sort of automatic network discovery.

I think it is important that no setup is needed. It should "just work" when the devices are in the same local network.
If people have to enter an IP-address, then it is too complicated.

A port scan could suffice for this.

**Network lib**
First, a network lib is required. This is also relevant for #68.

This [blog post](https://blogs.unity3d.com/2020/09/08/choosing-the-right-netcode-for-your-game/) names possible libs. Out of these, free open source solutions seem to be MLAPI and Mirror.

Note that MLAPI will be the starting point for Unity's upcoming [networking solution](https://blogs.unity3d.com/2020/12/03/accelerating-unitys-new-gameobjects-multiplayer-networking-framework/).

**Own project**

The companion app is a separate (GitHub) project.

The companion app is used to provide input for the main game and maybe control it (see also #176 for ideas).

The companion app has nothing to do with the multiplayer feature of #68 (except that the same networking lib should be used).

Example (PlantUML): 
```
@startuml
node "LAN1" {
  [Tablet]
  [Smartphone1]
  [Smartphone2]

  Tablet <--> Smartphone1
  Tablet <--> Smartphone2
}

node "LAN2" {
  [PC]
  [Smartphone3]
  [USBMic]

  PC <--> Smartphone3
  PC <--> USBMic
}

Tablet <--> PC
@enduml
``` 

- PC and Tablet are running the main game in different networks.
    - They receive mic input from smartphones, which run the companion app.
    - PC and Tablet communicate and exchange data for multiplayer. However, this is independent of the companion app.

**Names**
Any suggestions for names?
I would just name it "UltraStar Play Companion"

achimmihca:
### What does this PR do?

This is a major PR that implementes support for the Companion App (or similar clients) of #230.

- added SimpleHttpServerForUnity in Plugins to register REST endpoints 
    - I did not find such a ready-to-use solution, which is why I built this.
    - Uses C# standard HttpListener.
        - This might not be supported on all platforms (check HttpListener.IsSuported).However, I tested it with Windows and Android and it works fine there.
    - I built it such that it could be moved to a separate repo because it is not related to UltraStar.
    - Requests are handled on Unity's main thread for ease of use.
        - Of course this introduces a delay of several milliseconds. However, this is sufficient to fetch some data with a handful of clients.
        - The response time (main thread vs immediately) could be changed easily by a new parameter if needed.
    - paths can use parameters in curly braces. For example `api/rest/hello/{name}` would match for `api/rest/hello/bob` and use `bob` as value for the `name` variable
        - See CurlyBracePlaceholderMatcher (and its test). Uses Regex for implementation.
    - See UltraStarPlayHttpServer for an example

The HttpServer makes it easy to register endpoints and implement more features for #176.

- I had contact with few REST APIs before. I took the one from [JIRA](https://docs.atlassian.com/software/jira/docs/api/REST/8.1.0/) as inspiration
    - I suggest to use `/api/rest/` as prefix. In theory other APIs could be implemented (e.g. XML stuff)

- Communication is done by exchanging JSON messages
    - The data-classes use the suffix "Dto" for "data transfer object". These are simple plain-old-c#-objects (POCOs).
        - C# initializer syntax is quite nice to create a POCO with some data. Example `new MessageDto { theMessageField="hi" };`
    - Dto-classes are also used when answering REST requests of the HttpServer
        - Example: The currently loaded songs are sent via LoadedSongsDto

In the following, "companion app" is considered the "client" and the "main game" is considered the "server"
- ServerSideConnectRequestManager is listening for UdpBroadcast on some port (currently 34567)
    - The ClientSideConnectRequestManager is located in the companion app repo and performs such a broadcast 
- The connect request includes a parameter if a microphone port is desired. If this is set, then the server will create a new TcpListener to receive mic data.
    - This TcpListener port for the mic data is included in the response
    - The response also contains the HttpServer port such that the client can do REST requests.
        - ATM this is only done to query the list of loaded songs

- When a connection has been established, then the server checks if the client is still connected from time to time. I called this "still alive check". If the client cannot be reached anymore, then it is removed.
    - This is implemented by sending a single byte to the client.
    - The client reads this and throws it away. This is just to make sure that the socket on the client is not overflowing at some point.
        - In a similar way, more fancy messages could be sent if needed.
- The client sends its raw sample data to the main game.
- Amplification etc. is done by the main game.
- Assuming 44100 Hz, this would be 44100 floats. Assuming 32bit floats, this would be 44100 * 4 bytes, which is approx. 175 KB per second. This should be no problem for WLAN even with multiple connected clients.

I also found and fixed some issues
- Amplification was applied only to new samples in MicSampleRecorder. However, the mic buffer is overwritten completely by Unity when calling Microphone.GetData(). Thus, amplification has to be applied every frame.
    - I tested which is faster: (a) apply amplification to all samples every frame, or (b) apply amplification only to new samples and shift old samples in the buffer to make place for the new samples
        - I found that (a) is faster.
        - Also, I use *= for the multiplication and removed the "less-than -1, greather-than 1" check because it should not be needed. Pitch detection still works with some samples outside -1 , 1 range.
            - I found that this change makes it about twice as fast because less CPU instructions are needed for this operation.
            - For 10 * 44100 Hz, it takes about 3ms. So, this should be no problem, even with multiple clients. 

- Communication is handled in dedicated Threads.
    - Before I was using ThreadPool.cs as usual. However, ThreadPool is using only a single thread. Because the still-alive-check is sleeping for 250 ms, the whole ThreadPool was sleeping this duration. Using a dedicated Thread instead solves this.

- I found that the currentSamplePosition returned by Unity's Microphone.GetPosition does not change every frame (on Android). I guess Unity pulls this buffer from time to time on some platforms. Thus, there is a check now `lastSamplePosition != currentSamplePosition` before handling new mic samples
    - This was a nasty bug because there is little documentation from Unity.

- mini change: The mic-delay slider is now changing in steps of 10 ms, instead 1ms.
    - 10ms is still less than 1 frame and I imagine tapping the button 100 times on mobile for a significant change might be annoying.

### Closes Issue(s)

close #230 

### More

- I plan to create a release of the companion app soon.
    - also I want to integrate Travis and Codacy for the companion app repo.
- documentation follows after a new released has been created

### Additional Notes

- I already played a round using my smartphone as mic. It's awesome! :)
- I used a latency of 330ms for my companion app. Using USB mics I use about 250-300ms. So latency of the companion app is similar.


bohning:
Wow, @achimmihca . I can’t wait to try this!

achimmihca:
Find the Companion App here: https://github.com/UltraStar-Deluxe/PlayCompanion/releases/tag/v0.1

achimmihca:
codacy is complaining about thread variables that could be declared local.

However, I am unsure about that. The thread should not terminate. 
I would prefer having a variable to access the thread, just as reminder that it is still there.

basisbit (approved):

PR #237, opened 2021-04-08
Anst/improvements

achimmihca:
### What does this PR do?

This is PR fixes some issues and prepares the next release.

**Versioning**
- The BuildInfoGenerator adds the commit hash of git to the VERSION.txt file (e.g. `5e07588`). And the MainScene shows this next to the regular semantic version number.
    - I realized that some automatic mechanism is needed to tell builds apart. The previous build-timestamp approach is not optimal, because the same code base could be build at a different time. The commit hash is a better automatic identifier for the version.
- The regular semantic version number in VERSION.txt is now taken from the ProjectSettings.
    - This way, only one version must be adjusted.
    - ProjectSettings.bundleVersion is probably more important because this could affect other places, such as Google Play Store, App Store, etc. Thus, this value is copied to VERSION.txt and not vice versa

**Refactoring**
- Many files changed because I removed `List<IDisposable>` and instead use UniRx' extension method AddTo(GameObject)
    - The extension method calls Dispose() in OnDestroy of the GameObject, just like before

**fixes**
- The (dis)connect event of a companion app was not handled in some places (e.g. SongSelect). This has been improved.


achimmihca:
@basisbit  any news on this PR?

basisbit (approved):

PR #244, opened 2021-05-02
Anst/fixes

achimmihca:
### What does this PR do?

A minor PR that fixes tons of small issues:

- only show available languages in language slider
    - No need to show languages that fall back to English anyway
- Player profile image was not correctly shown in sing-scene and singing-results-scene
    - was always the default image
- lyrics when using ScrollingNoteStream were *behind* their background
    - This made the lyrics more difficult to read instead of easier to read
- some duet songs were not working. Was using wrong "voice name"
    - in song files, it is possible to provide names for `#P1` and `#P2`. When these were present, it crashed
- cannot select next/previous song when searching
    - no idea why I added this condition

Also fixes tons of issues with the song-editor:
- context menu of sentence in song-editor not working
    - the Text component was in front of the raycast target, blocking the context menu
- sentence should be removed when all its notes are removed
- faster changing playback position in increments (hold Ctrl and right/left arrow)
    - 1ms is nothing. I increased it to 10ms.
- assigning new notes to voice was choosing the wrong voice
- "Move notes to player" and "paste notes" should create a single sentence for the notes (instead of one sentence per note)
- new "move to own sentence" action
- wrong font size on notes sometimes
- fix NPE
- dont add space as lyrics to newly created note when recording notes

Fixed conflicting controls in song-editor:
- "space" in TextArea should not start/stop playback
- "delete" or "select all" in TextArea should not affect selected notes
- "escape" in TextArea should remove focus from TextArea, not exit the scene
- "escape" with content-to-be-pasted should cancel content-to-be-pasted, not exit the scene
- "escape" with selected notes should deselect notes, not exit the scene

Will create a new 0.3.1 release when this has been merged

### DIFF
--- a/UltraStar Play/Assets/Common/I18N/I18NManager.cs
+++ b/UltraStar Play/Assets/Common/I18N/I18NManager.cs
@@ -218,4 +218,21 @@ public class I18NManager : MonoBehaviour, INeedInjection
         path = ApplicationUtils.GetStreamingAssetsPath(path);
         return path;
     }
+    
+    public List<SystemLanguage> GetTranslatedLanguages()
+    {
+        HashSet<SystemLanguage> translatedLanguages = new HashSet<SystemLanguage> { SystemLanguage.English };
+        foreach (SystemLanguage systemLanguage in EnumUtils.GetValuesAsList<SystemLanguage>())
+        {
+            string suffix = GetCountryCodeSuffixForPropertiesFile(systemLanguage);
+            string propertiesFilePath = GetPropertiesFilePath(PropertiesFileName + suffix);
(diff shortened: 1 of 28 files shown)


basisbit (approved):

basisbit:
There are a couple of places where exceptions are handled by just ignoring faulty states instead of at least logging them. That should be improved in the future to make it easier to actually find and resolve problems on users devices which are hard to reproduce...

achimmihca:
I just searched our code for ignored exceptions without logging.
I did not find any in our code.
I only found few inside "Plugins" folder, i.e., in some libs that we use.

Do you have an example in our code?

PR #251, opened 2021-09-23
Serilog-Dependency correction

supernevi:
there is no master-branch anymore. only dev and main

commit hash stayes the same


basisbit (approved):
lgtm. 👍 
Thank you for this contribution! 🚀

PR #258, opened 2021-10-25
Anst/song editor improvements

achimmihca:
### What does this PR do?

This is a small PR that implements a different way to stretch/shrink a selection of notes in the SongEditor.
- Before, the drag distance was just added to the EndBeat/StartBeat.
    - By holding shift, it is now possible to scale the selection of notes, such that they keep the distances relative to each other (plus/minus rounding issues because of beat grid size).

Besides this,
- The SongEditor now opens where the lyrics begin by default.
- The SongEditor "auto scroll border" is smaller now to reduce the number of jumps and make the editing area larger.
- I commented SongMeta.cs because I always forget the units of GAP and VIDEOGAP

### Motivation

I am editing songs and wanted this feature.


basisbit (approved):
lgtm

basisbit:
I like the new Github Actions integration a lot!

Issue #231, opened 2021-03-14
Migration to UIToolkit


achimmihca:
Most of the UI is currently implemented using a Component-based "Unity UI" (aka. uUI).

[UIToolkit](https://docs.unity3d.com/2020.2/Documentation/Manual/UIElements.html), which should be released in a stable version in 2021, will provide better separation of concerns and better maintanability.
Therefor, UIToolkit builds on XML and Stylesheets, very similar to HTML and CSS.

I think the new flex-layout engine and stylesheets will make UI stuff much easier and consistent to implement.
Furthermore, having the UI structure in XML makes tracing and understanding changes easier in version control.

### Missing features
UIToolkit is still under development and some features that we use with uUI are missing.

- [ ] UI Toolkit does not yet provide a way to load a style from a string ([forum](https://forum.unity.com/threads/generate-visualelementasset-stylesheet-from-a-string.851800/)).
    - This is required to load user defined themes. Once this is supported by Unity, theming will be trivial to implement: add the loaded StyleSheet to the UIDocument's StyleSheet list.

- [ ] UI Toolkit does not yet provide a way to use custom shaders on images
    - This is required to implement the HSV shader that is used to change color of some images (e.g. player avatar image)  ([forum](https://forum.unity.com/threads/change-image-color-in-hsv-range.1030366/)).
    - Maybe an alternative would be to change the pixels of the Texture2D?

- [ ] Rich Text Support (change colors, bold and italic embedded in the text)
    - Seems to be available in 2021.2 (currently available only as pre-release) ([forum](https://forum.unity.com/threads/ui-builder-latest-version-1-0-0-preview-12.823545/#post-6564745))

### Migration Status

- [x] MainScene is using UIToolkit already as a first test (see #220).
- [ ] Options
    - The controls have to be implemented using UIToolkit (number selection (e.g. mic delay), enum selection, etc.) 
    - The RecordingOptions are visualizing the mic input by drawing on a texture.
        - `DynamicTexture.cs` has been prepared to do the same using UIToolkit.
   - Draws player images in different colors
- [ ] SongSelect
    - Uses drag-handlers. Should be similar in UIToolkit ([documentation](https://docs.unity3d.com/2020.2/Documentation/Manual/UIE-Events-DragAndDrop.html)).
    - Visualizes mic input (like the RecordingOptions).
-  [ ] SingScene
[...]

Issue #68, opened 2019-11-29
online multiplayer mode


basisbit:
Issue type: Feature request

### Actual behaviour
Currently UltraStar Play does not support playing karaoke together with friends who are not in the same room/location.

### Expected behaviour
The game should support some multiplayer mode where users can use their own device or share a device to and play together with users of other devices running UltraStar Play.

### Topics to consider
- [ ] how can different devices find each other? maybe some online matchmaking / party finder website that could be hosted on usplay.net
- [ ] we probably don't want to rely on commercial-only Unity online services.
- [ ] consider using a network multiplayer framework that supports various platforms that Unity supports. Maybe something like this: https://github.com/MidLevel/MLAPI
- [ ] local multiplayer might be nice eventually
- [ ] decide on what development path to take. The first iteration of the multiplayer feature should only contain what would usually be considered the minimum viable product (MVP). Once that is implemented, more developers will probably help developing the fancy features.
- [ ] people have different song file databases with different variations of song txt files - how should matchmaking work and still be fair? For legal reasons, the online multiplayer mode may not exchange / transfer / store any lyrics or song txt files or similar. Maybe a md5/sha1 hash of the song txt files minus the headers could be used as song "identifier" and to support party match making.

achimmihca:
This [blog post](https://blogs.unity3d.com/2020/09/08/choosing-the-right-netcode-for-your-game/) names possible libs. Out of these, free open source solutions seem to be MLAPI and Mirror.

MLAPI will be the starting point for Unity's upcoming networking solution.
They just released some [docs](https://docs-multiplayer.unity3d.com/docs/getting-started/about-mlapi).

I already tried the current state of MLAPI in a branch (anst/mlapi). It is quite nice to implement mulitplayer features.
It supports all that is needed for USPlay.I think implementing multiplayer this way could be relatively simple.
MLAPI should be included in Unity's next major release.

So I suggest to wait until then before digging deeper.

bohning:
> * people have different song file databases with different variations of song txt files - how should matchmaking work and still be fair? For legal reasons, the online multiplayer mode may not exchange / transfer / store any lyrics or song txt files or similar. Maybe a md5/sha1 hash of the song txt files minus the headers could be used as song "identifier" and to support party match making.

Regarding this point, maybe some kind of "song normalization" (performed internally, without touching the actual text file, or only with user consent) could be performed. Meaning, e.g.

[...]

PR #259, opened 2021-10-29
upgrade to Unity 2021.2.0f1

achimmihca:
### What does this PR do?

Upgrade to Unity 2021.2.0f1

Unity has finally released the new major version.
Noteworthy changes include
- the new UIToolkit as a regular package (for #231)
- the new networking API as a regular package (for #68)

I also updated the GitHub Actions Unity version and the UNITY_LICENSE environment variable in GitHub.
Let's see if this "just works".


basisbit (commented):
lgtm. Thanks for the PR! 🥳

PR #260, opened 2021-10-29
build macOS to check if  compilation works using unity-test-runner

achimmihca:
### What does this PR do?

For some reason the macOS executable build does not work with unity-builder GitHub Action.
This PR is to check, if the compilation also fails with the unity-test-runner.

Please, don't merge this PR.


gianpaj:
still in draft?

achimmihca:
The tests work fine on OS X using unity-test-runner.
However, the build on OS X still fails.
Don't know why, might be fixed with a new version of Unity or unity-builder GitHub Action.

PR #268, opened 2021-12-08
Anst/more UI toolkit

achimmihca:
### What does this PR do?

Implements HighscoreScene and SingingResultsScene using UIToolkit.

- SingingResultsScene hides certain elements (e.g. player image) when there are lots of player results to display
    - This is done using StyleSheets. See MainStyles for details
- Statistics and graph in SingingResultsScene are not done yet. Future work...
- Sadly, the playerImage colors look stupid now, because UIToolkit's color property multiplies the colors (i.e. it gets darker). This will be fixed easily with the HSV color shader once UIToolkit supports shaders (hopefully next year).
- SingingResultsScene works for up to 9 players this way (more or less beatiful).
    - Custom layout for 3, 4, 5, 6 players (like it has been done in USDX) would be something for future work

### More

SongSelectScene will be next :)


basisbit:
can you please create issues to track the new tasks that need to be done in these scenes?

Issue #166, opened 2020-04-14
Log and skip unknown file extensions


bohning:
This idea came out of a discussion in PR #164. When scanning the song folder(s), files with unknown file extensions could be logged and skipped.

achimmihca:
Note that this should be done for audio files as well as video files.
If a song uses an unsupported video format, it should be logged. This makes it easier to migrate a whole library of songs to a supported format.

Baklap4:
I wouldn't rely on the extension as it can be faked and even wrong. Not sure if unity keeps track of this but i'd suggest reading the first few bytes, and check the file headers to determine if it's a file you can process or not

bohning:
> I wouldn't rely on the extension as it can be faked and even wrong. Not sure if unity keeps track of this but i'd suggest reading the first few bytes, and check the file headers to determine if it's a file you can process or not

I would consider this extra credit. In a first version, considering file extensions and logging unsupported files would be of great benefit already (and pretty simple to implement, I suppose).

achimmihca:
> I wouldn't rely on the extension as it can be faked and even wrong

I would not bother detecting such issues while scanning for songs. And I would not start implementing media decoding. Unity is the platform on which USPlay is built, so we should rely on their API and media support.
If Unity API has issues opening a file, then we should handle it as it occurs. But in-depth analysis of what might and might not work beforehand is wasted effort in my opinion.

basisbit:
in general, the idea is to get much more people to just download/install usdx format tar/zip files and feed them to the game. Thus, people who create those song packages usually make sure that the content works. Especially on smartphones or tablets or TVs you'll not want to deal with file formats or adjusting song timing.
All in all this will hopefully improve the song files quality, get people to use modern + widely supported file formats (see [here](https://github.com/UltraStar-Deluxe/Play/wiki/Converting-Song-Media-Files), and make it much easier for occasional players to add their content to the game.

PR #277, opened 2022-02-13
Changed default encoding to UTF8

### LINKED ISSUE #276: [Mac] Play can’t find songs that have a typographer’s apostrophe in it
Issue type: Bug report

I try to use the typographer’s apostrophe (’) instead of the typewriter’s apostrophe (https://en.wikipedia.org/wiki/Apostrophe#Typographic_form) in my filename. I added a task to UltraStar Manager that can replace all wrong uses (accents:  `´) and typewriter’s apostrophes by the (more correct) typographer’s apostrophe. The typographer’s apostrophe is also included in Windows CP1252, so even old UltraStar (before 1.1, also CMD) can handle this character.

When I start Play with my song collection I noticed that about 20 % are missing when compared to USDX and after a while I noticed that all the songs with apostrophe’s in their filenames are missing. The problem with the apostrophe also becomes evident when starting songs that have apostrophes in their lyrics (basically all English songs):

image

So somehow the typographer’s apostrophe cannot be decoded and is displayed as "???". So then I started play with only one song that had a typographer’s apostrophe in its filenames - and indeed, Play says "No songs found".

Looking into the log file, I find:

image

Meanwhile, USDX does not have a problem with the file at all, so the filenames referenced in the text file and the filenames themselves are indeed identical (and must be, otherwise USDX would complain as well).

image

This is where it gets strange: the same file works on Windows without problems (tested by @achimmihca).

### Steps to reproduce
Start Play with a song that has a typographer’s apostrophe in the filenames.

### Details
Provide some additional information:

- UltraStar Play version: latest master
- Operating System + version: macOS 12.1 Monterey
- if applicable, add a log file

Relevant lines from the log are:

```
2022-02-06 21:29:41.271 [WRN] Audio file does not exist: /Volumes/Karaoke/++/SongsHomeGrown/Beatsteaks - Mietzi’s Song [MB&HR]/Beatsteaks - Mietzi???s Song.mp3
2022-02-06 21:29:41.271 [INF] Finished song-scan-thread after 22 ms. Loaded 0 songs. Failed 1 songs.
```

Does anybody have an idea what’s going wrong?

achimmihca:
### What does this PR do?

- Changed default encoding from `Encoding.Default` to UTF8 in TxtReader.GuessFileEncoding
    - `Encoding.Default` seems to be system dependent
    - UTF8 with and without BOM should now work on all platforms.
    - Further, UTF16-BE with BOM and UTF16-LE with BOM should still work on all platforms.
    - Files with ANSI encoding might be misinterpreted now. Solution: convert them to UTF8.
- Add test that reads txt files with different encodings
- Also: removed unused methods from TxtReader

I strongly vote to use UTF8 as default encoding instead of ANSI.
I expected this to be the case already. However, `Encoding.Default` was used as default, which is system dependent.

### Closes Issue(s)

close #276


UltraStar Play/Assets/Editor/Tests/TestSongs/TestSong-ASCII.txt
```diff
@@ -0,0 +1,11 @@
+#TITLE:TestTitle___�___�___'___?
```
achimmihca, line 1: Looks like GitHub is also interpreting the file as UTF-8

bohning:
Thanks for taking the time to look into this, @achimmihca !

bohning (approved):

bohning:
I support your vote for UTF-8. However, I have a feeling there is a non-negligible user base of the old USDX (Windows-only), which does not support UTF-8 (only CP1252).

I can confirm that your PR works as expected - regarding my test of different encodings, Play now finds:
- UTF8-noBOM
- UTF8-BOM
- UTF16-BE-BOM
- UTF16-LE-BOM

All songs in my song collection (all UTF-8 encoded) are now found successfully (#276) on macOS.

basisbit:
Those non-UTF8 Code Pages which just extend ASCII have always been a pita for karaoke game development. Referring to the message in https://github.com/UltraStar-Deluxe/Play/issues/166#issuecomment-628204639 , pushing people to use the package format should also help get rid of ANSI Codepage encoded txt/xml files.

bohning:
@basisbit Is there a specification for the "package format"? Is it just a zip/tar file of the song files? Including the song folder or without it? This package can then be read by Play and the files are extracted on the fly?

basisbit:
it is just a tar file of a folder that contains the songs content. The plan was to also support zip in the future (there was a licensing issue with SharpZipLib, which should be resolved by now). Currently a download url can be pasted in the usplay content download scene. That then unpacks the content to a local app data folder. This could be extended to support kind of a repository lookup, where the user only specifies a repository url, the game then loads a content list and asks the user what packages should be downloaded.

achimmihca:
> The plan was to also support zip in the future

Has been implemented by now

Issue #246, opened 2021-05-22
Problem playing mp4 music videos on ubuntu 18.04LTS


mnissov:
Issue type: Bug report

### Actual behaviour
In sing mode when going over a song with a video file a red error message appears saying "Can't play movie [file:///home/user//...]. This happens with both .avi and .mp4 video formats.

The same song repository works with ultra star deluxe.

### Details
Provide some additional information:

- UltraStar Play version: 0.3.1
- Ubuntu 18.04 LTS

basisbit:
Please add more information, like file format and so on. UltraStar Play only supports file formats which Unity framework can play on your OS, plus mp3 for which we added additional support.

mnissov:
> Please add more information, like file format and so on. UltraStar Play only supports file formats which Unity framework can play on your OS, plus mp3 for which we added additional support.

Edited the original, but restated it happens with .avi and .mp4. Not sure if my case has a working video format, all the videos I have are in these formats.

paradigmn:
According to the Unity [manual](https://docs.unity3d.com/2021.1/Documentation/Manual/VideoSources-FileCompatibility.html) only .ogv .webm and .vp8 videos are supported on linux. I made a couple of tests and the only thing that worked for me was a VP8 encoded video with WebM container. The file had to be converted with [ffmpeg](https://trac.ffmpeg.org/wiki/Encode/VP8).

achimmihca:
This surprises me, but you are right. A recent (Feb 2021) [forum post](https://forum.unity.com/threads/error-importing-mp4-file.711203/)

> At this time we don't support MP4 import on Linux. ... H.264 / H.265 are patent-encumbered. With Mac and Windows this is less of an issue since the platforms themselves cover licensing. With Linux the story is murky since no distro does this.
I like the idea of supporting 3rd party codecs at some point. We don't have the bandwidth right now, though.

So mp4 is not supported on Linux. And it seems that adding this codec via Unity API is not supported either.

This sucks, I expected mp4 to work on every platform. Maybe there is some other C# lib that we could use.
There are C# wrappers for ffmpeg: [StackOverflow](https://stackoverflow.com/questions/6717184/net-ffmpeg-wrapper-for-video-playback), [MediaToolkit.NetCore](https://github.com/mtebenev/MediaToolkit.NetCore).
Not sure if these will work with Unity on Linux. Note that we would need a lib that is compatible with our MIT license.

basisbit:
[...]

PR #278, opened 2022-02-22
Anst/stream media

achimmihca:
### What does this PR do?

- allow the use of http and https URIs as audio, video, and image in song files.
- fix performance issue in the SongSelectScene.
    - every audio file was loaded, just to show its duration. This is why scrolling was stuttering.

**Streaming media files**
For example, consider [Joshua Morin - On the run](https://github.com/UltraStar-Deluxe/songs/tree/master/Creative%20Commons/Joshua%20Morin%20-%20On%20the%20run).
This UltraStar song can now be written as:
```
#TITLE:On the run
#ARTIST:Joshua Morin
#EDITION:Creative Commons
#LANGUAGE:English
#MP3:https://raw.githubusercontent.com/UltraStar-Deluxe/songs/master/Creative%20Commons/Joshua%20Morin%20-%20On%20the%20run/audio.ogg
#COVER:https://raw.githubusercontent.com/UltraStar-Deluxe/songs/master/Creative%20Commons/Joshua%20Morin%20-%20On%20the%20run/cover.jpg
#BACKGROUND:https://raw.githubusercontent.com/UltraStar-Deluxe/songs/master/Creative%20Commons/Joshua%20Morin%20-%20On%20the%20run/background.jpg
#VIDEO:http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4
#BPM:297,5
#GAP:11250
: 0 3 9 So
: 6 4 11  far
: 12 2 9  a
: 16 4 11 way
* 25 8 14  etc.
- 44
E
```

Note that the URI must point directly to the media file (YouTube links won't work).

**Supported formats**
The supported formats are the ones that Unity supports natively for streaming.
This is platform dependent, just like before . As a rule of thumb:
- ogg should work on all platforms.
- mp3 does not work.
    - We use an external lib to play mp3 files
    - One could download the bytes upfront. But this would be a long running download (in contrast, ogg is streamed on-the-fly by Unity). Thus, I argue to focus on ogg.
- mp4 does not work on Linux (see #246). But it should work on Windows and Mac. It might also work on Android and iOS, not sure.
- VP8 as video should work on all platforms, but might be slow if there is no hardware support.

**File extension**
The URI must end with the file extension of the used format.
For example, an ogg audio URI must end with `.ogg` and a VP8 video URI must end with `.vp8`

### Motivation

- Beeing able to stream the media files makes the song access/distribution easier, especially on mobile platforms.
- It should be fairly simple now to create a song repo.
    - One could add some authentication to the media requests (e.g. via HTTP basic auth in the request's header). This could be used to allow access only for authenticated users, e.g., to a private repository or commercial song service. ( FYI @bohning  )

### Additional Notes

This was surprisingly simple. Would have done it earlier if I knew this :)

### DIFF
--- a/UltraStar Play/Assets/Common/Audio/AudioManager.cs
+++ b/UltraStar Play/Assets/Common/Audio/AudioManager.cs
@@ -49,10 +49,15 @@ public class AudioManager : MonoBehaviour
         }
     }
 
-    // When streamAudio is false, all audio data is loaded at once in a blocking way.
-    public AudioClip LoadAudioClip(string path, bool streamAudio = true)
+    public AudioClip LoadAudioClipFromFile(string path, bool streamAudio = true)
     {
-        if (audioClipCache.TryGetValue(path, out CachedAudioClip cachedAudioClip)
+        return LoadAudioClip("file://" + path, streamAudio);
+    }
+
+    // When streamAudio is false, all audio data is loaded at once in a blocking way.
+    public AudioClip LoadAudioClip(string uri, bool streamAudio = true)
+    {
+        if (audioClipCache.TryGetValue(uri, out CachedAudioClip cachedAudioClip)
             && (cachedAudioClip.StreamedAudioClip != null || cachedAudioClip.FullAudioClip))
         {
             if (streamAudio && cachedAudioClip.StreamedAudioClip != null)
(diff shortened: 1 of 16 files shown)


basisbit:
please take a look at the test failures (TestCamdPitchDetection fails for some reason). Thanks!

basisbit (approved):
Lgtm. Thank you for the pull request! 🚀

PR #285, opened 2022-03-30
refactoring: float instead double

achimmihca:
### What does this PR do?

Use float instead of double everywhere.

Motivation:
- Unity API uses float most of the time (e.g. coordinates of UIElements)
- number range of float is pretty much always sufficient
    - first integer not representable in a float is 16,777,217 (when used as milliseconds, this would be more than 4 hours)
    - for percentages and usual number ranges, the fractional precision is also sufficient
- casting between double/float is annoying
- mixing float and double has neglectable better precision compared to only using float
- float is slightly more efficient
    - given that some stuff is executed 60 times per second, and also on mobile devices, this could be a noticable difference


basisbit (commented):
I suggest not merging this.
- float is not faster than double on typical current systems
- float is very much not precise, that has already caused multiple of bugs/issues in UltraStar Deluxe as well as Vocaluxe and Performous

achimmihca:
>  that has already caused multiple of bugs/issues in UltraStar Deluxe as well as Vocaluxe and Performous

I did not expect that.

> I suggest not merging this.

Ok, I can live with that.

> instead put in the effort to check what the code is doing

Not worth it.

PR #287, opened 2022-03-31
Anst/refactoring object instantiation

### LINKED ISSUE #283: Code Style: remove redundant type in instantiation (`Bla bla = new Bla()` vs. `Bla bla = new()`)
### Problem or question to be solved
We once decided against type inference using `var` keyword because we wanted to specify the **resulting** type explicitly.
Thus, we currently use the redundant form `Bla bla = new Bla()` instead `var bla = new Bla()`.
This is especially noisy with generics, e.g. `Dictionary<string, string> string2string = new Dictionary<string, string>();`

### Suggested solution/s
However, newer C# versions allow to skip the redundant constructor type.
Thus, one could write `Bla bla = new()`.
Note that the resulting type is still explicitly specified.

### Your opinion on the topic
I think this is a smart c# feature.
It removes the redundant type specification but on the right-hand-side, such that the resulting type is still clear for the reader.

Thus, I argue that we should use this. My IDE has a nice "code cleaup" feature, that can apply this to whole code base.
Any other opinions ( @basisbit ) ?

achimmihca:
## What does this PR do?

Remove redundant type in instantiation as discussed in #283
This time, without previous "double -> float" refactoring.

## Closes Issue(s)

close #283

### DIFF
--- a/UltraStar Play/Assets/Common/ApplicationManager.cs
+++ b/UltraStar Play/Assets/Common/ApplicationManager.cs
@@ -17,7 +17,7 @@ public class ApplicationManager : MonoBehaviour, INeedInjection, IInjectionFinis
         }
     }
 
-    public List<string> simulatedCommandLineArguments = new List<string>();
+    public List<string> simulatedCommandLineArguments = new();
 
     [Range(-1, 60)]
     public int targetFrameRate = 30;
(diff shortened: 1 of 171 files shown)


basisbit (approved):
looks good. Thank you for this pull request and the huge amounts and speed of contributions! 🚀

PR #288, opened 2022-04-03
Anst/refactoring

achimmihca:
### What does this PR do?

**Mostly refactoring:**
- remove unused assets
- move some files
    - notably Common/UIToolkit has been merged with Common/UI, because there is only UIToolkit now
- Removed serializable collections sub-classes. Unity by now supports serialization of generic objects

**Small fixes:**
- AvatarImageControl did not show initial image
- hide theme picker in options
    - because the old theming approach has been removed
- SingScene was showing the same lyrics twice in some cases
    - example case: first and second player use Voice1 and third player use Voice2
    - the selected voice of the player control must be considered when choosing one for the lyrics control

### More

- This is the last refactoring before creating a common code base for main game and companion app


basisbit (approved):
Thanks for the pull request, looks good!

PR #290, opened 2022-04-06
Anst/shared repo

achimmihca:
### What does this PR do?

I created a [repository for common code](https://github.com/UltraStar-Deluxe/PlaySharedRepo) of the main game and companion app.
This PR reflects the changed on the main game.
- Moved files mostly
    - and changed paths in UXML files
- Few times, I needed to add an interface that main game and companion app implement in different ways (notably `IConnectedClientHandler`)

The `SharedRepo` is located in `Assets/Common/SharedRepo` as a [git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules), which is basically one Git repository inside another.

One can pull / push changes from / to the SharedRepo as usual.
Git stores the exact commit of the SharedRepo that is currently used.
Thus, pushing to the SharedRepo does not affect main game or companion app without explicitly pulling these changes.

### Additional Notes

- Now, the companion app can access the same code as the main game ( https://github.com/UltraStar-Deluxe/PlayCompanion/issues/1 )
- Next, I want to implement pitch detection on the companion app ( https://github.com/UltraStar-Deluxe/PlayCompanion/issues/3 ).
- Afterwards I want to create a new release and then focus on YouTube and tutorials for a while.


basisbit:
Have you considered keeping the shared code here in this /Play repository, instead of in yet another repository?
Splitting so much code which is crucial to the project out of the project's main repository and its ci/cd workflow feels kind of unintuitive to me, tbh,

Depending how strongly the companion app depends on the code her ein this repo, it might be worth it to have the code from the companion app and from the main game together here in the /Play repository, similar to how Vocaluxe does it - see VocaluxeLib in https://github.com/vocaluxe/Vocaluxe/

(I don't have made up a personal opinion on this yet, this is just asking for some comments and some discussion, to make sure that this works well in the long run and is easy to refactor / debug / edit)

achimmihca:
I agree, it is very cumbersome.
A single repo would be nice, will try some stuff.

achimmihca:
I managed to have both Unity projects in a single repo.
- The `SharedRepo` folder can be added as symlink in the companion app (i.e. SharedRepo folder of companion app points to SharedRepo folder of main game).
    - I created a shell script for this that should work cross-platform
        - I tested it on Windows, the GitHub Action will test it on Linux, MacOS will probably work if Linux does
        - One must run the script initially to set up a new development environment (just like the download-dependencies script)
        - Maybe one could create a single script `setup.sh`, which runs both, download-dependencies.sh and the symlink script.
    - The symlink can be added to .gitignore. The files are always committed from their "real" location.
- The GitHub Actions to run tests and build the companion app should be easy to add as well
    - Each PR would run tests of main game and companion app
    - Similarly, main game and companion app would always both be built when pushing to main branch
    - Advantage: changes to files in SharedRepo would always be tested on main game and companion app

@basisbit , what do you think?
I think a single repo would make management easier, also considering issues on GitHub etc.

achimmihca:
Symlinks are not the way to go 😨 !
Today Unity crashed and I got a warning that symlinks may corrupt a Unity project. Seems like it messes up the `.meta` files.

But then I thought about other ways to share assets in Unity projects. And thankfully there is Unity's package system that is already used to include some libraries (UniInject, ProTrans, etc.)

Thus, I move the SharedRepo files to its own Unity package located inside the main game's package folder.

This package can then be referenced by the companion app in `manifest.json`:
```
{
  "playshared": "file:../../UltraStar Play/Packages/PlayShared",`
  ...
}
```
Note that I renamed if from "SharedRepo" to "PlayShared".

In conclusion, this is the way to go:
- A single repo for everything (aka. monorepo) for easier management of issues, wiki, code, reviews, everything.
- Sharing common assets between Unity projects as Unity package
    - No git submodule, git subtree, symlinks or other tools required.
    - Because it is its own package and assembly, it also enforces proper direction of dependencies
        - For example, interfaces might be needed to abstract common stuff of classes

Will open a new PR.

PR #294, opened 2022-05-06
Anst/pitch tracking on app

achimmihca:
### What does this PR do?

**Pitch Analysis on Companion App**
- See (https://github.com/UltraStar-Deluxe/PlayCompanion/issues/3)
- Only the pitch analysis result is sent to the main game.
    - Pitch analysis result is tuple of (MidiNote, AnalyzedBeat)
        - If not in a song (e.g. in RecordingOptionsScene), then the newest samples are analyzed and AnalyzedBeat is -1
    - Previously, the mic samples were sent to the main game, which was waaay too much data
- The pitch analysis is now distributed. Thus, the main game may be run on a low-end device together with many smartphones, each doing their own pitch analysis.
- To do pitch analysis the position in song, BPM, and GAP are sent to the client
    - This is done frequently to stay synchronized (approx. every 2 seconds)
    - The companion app keeps track of the milliseconds since last time that it has received the position in the song
    - The companion app implements a mechanism to ignore single messages with big deviation. This makes the received position more resilient to jitter of network / OS / firewall etc.
        - I experienced delayed messages from time to time, i.e. position in song was more than 1 second later than expected. Such outliers are ignored now.
        - As a result, multiple messages need to be sent when doing an irregular position change in the song, e.g. because of "skip to next lyrics"
- Main game ignores pitch results for already analyzed beats and future beats
- Main game can handle multiple pitch results at once (e.g. when companion app received a new position in song)

- Corresponding refactorings
    - What was the "MicrophonePort" (to send mic samples) is now the "MessagingPort" (to exchange JSON messages)
    - ClientSideMicSampleRecorder has been removed. The normal MicSampleRecorder works perfectly fine.
        - This is now possible because main game and companion app have a common code base.
    - Better symmetry of main game and companion app messaging classes (IClientSideConnectRequestHandler, IServerSideConnectRequestHandler, IConnectedServerHandler, IConnectedClientHandler, etc.)

**New Companion App UI**
- Hamburger button to open menu on the side
- The song list can be filtered (https://github.com/UltraStar-Deluxe/PlayCompanion/issues/2)
- Companion app uses same translations as the main game
- Prevent stand-by when recording

**Singing UI Changes**
- Lines are not drawn. They don't add to the game and distract from the nice videos / backgrounds IMO.
- Notes are drawn as outline. These are filled with recorded notes.
    - I took inspiration from "Let's Sing 2019", which is a commercial singing game just like SingStar.
- Changed and added some mic profile colors to be more vivid
[...]

PR #298, opened 2022-05-09
unity: settings: enable runInBackground

bitnimble:
### What does this PR do?

This simply enables the `runInBackground` option for Unity, so that the game doesn't pause when it loses focus. This makes it have a sane behaviour that matches the original UltraStar as well.

### Motivation

I like to sometimes put UltraStar on my second monitor and sing along while doing other things, or briefly alt-tab to another window while a song is playing. Currently the game completely halts due to the Unity default settings and it's very jarring.

(If this is desired functionality, can we get it as an in-game setting instead?)

### Additional Notes

Unity docs link: https://docs.unity3d.com/ScriptReference/Application-runInBackground.html

### DIFF
--- a/UltraStar Play/ProjectSettings/ProjectSettings.asset
+++ b/UltraStar Play/ProjectSettings/ProjectSettings.asset
@@ -76,7 +76,7 @@ PlayerSettings:
   androidFullscreenMode: 1
   defaultIsNativeResolution: 1
   macRetinaSupport: 1
-  runInBackground: 0
+  runInBackground: 1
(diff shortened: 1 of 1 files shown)


achimmihca:
Thanks for pointing this out, I agree it is cumbersome.

In fact, this change is also included in #294 because I needed to move back and forth between the main game and companion app during testing.

I just forgot to mention that this setting will be changed as part of the other PR.

bitnimble:
Ah perfect! I'll close this one then 👍

PR #314, opened 2022-06-22
Anst/improvements

### LINKED ISSUE #279: Co-op modes for combined scoring
### Actual behaviour

Each player has his/her own score. This is also true for duet songs.

### Expected behaviour

There should be an option to combine scores when playing with multiple players.
This gives a co-operative touch to the game instead of a competitive one.
Especially for duets, I think this would make sense, because both players are singing together, not against each other.

I suggest two different modes:

**co-op mode 1: average score**
- A single score is shown for all players
- The score is calculated as the average score of all players
    - This does not make it easier to get a high score, but is feels more co-operative
    - Example: If all players have 10000 points, then the result is 10000 points. If one player has 0 and the other has 10000, then the overall score is 5000.
- Note that the score of each individual player is calculated as before. Only the displayed differs.

**co-op mode 2: take best**
- A single score is shown for all players
- The score is calculated as the average score of all players
- BUT: a beat of a player is considered as correct if it is considered as correct for **any** player
    - This makes it significantly easier to get a high score, because only the "best" of all players is considered.

- co-op mode 1 is a nice co-operative twist for duets and in general for players who want to sing together (instead of against each other), but who do not want to get an "easier" score.
- co-op mode 2 is nice for players, who need some hand-holding or do not want to impact the score negatively (yes, it's just a game but maybe someone is coy).
- Players may switch to the "average score" mode if the "take best" mode is too easy (which could be, needs testing).

**I'd like your opinion**
- I am sure that co-op mode 1 (average score) would be nice for the game.
- However, I am not sure if the "take best" mode is needed. What do you think?
- How should the scores be recorded?
    - At the moment, a score is recorded individually for each player and difficulty
    - But each player in co-op mode might have a different difficulty
    - I think a new "co-op score record" is required.
        - It stores the final score and for each player the name and difficulty.
        - This can be displayed on a dedicated co-op highscore list (so you would have: easy, medium, hard, co-op).
            - Difficulty of each player is shown, but it has no impact on the order in the list (only the final score is relevant).
    - For simplicity, the co-op score might not be recorded at all (in a first version)

### LINKED ISSUE #296: Add background music
### Actual behaviour

No music is playing in the menu.
Feels a little bit sad.

### Expected behaviour

Some background music should play when in the menu. This is common behavior in games and USDX has it as well.
Of course, the background music should not play in the song select, sing scene, and song editor.
The background music in the menu can be turned off in the settings.

I suggest to reuse the background music from USDX.
Or is there any other background music that we could use?

### LINKED ISSUE #305: View and copy log in-game
### Actual behaviour
With a growing number of users (yay!), there are more issue reports from various platforms.
The log is hidden somewhere in the file system and on mobile not easily accessible.

### Expected behaviour
I think it would make sense to add a TextField with the log somewhere in the development options (or accessible when developer mode is active on companion app).

For the implementation, one could
- add a new Serilog sink that logs to a list of log entries (log entry = message + severity)
    - it keeps only the last 1000 entries
- provide a TextField and a severity filter in dev options. By default it shows "warning and above".
    - it should be possible to copy the TextField content, such that users can paste the errors on GitHub issues.

Note that the log file may contain log entries from previous runs on the same day (log file changes by date and file size).

Further, the path to the log file should be shown in the developer options because it is a different path on every platform

achimmihca:
### What does this PR do?

- show log in-game in development options
- option for common score
    - Shows the average score of all players. The average score result is also saved as highscore with combined player names.
    - Example: Alice has 2000 points, Bob has 4000 points. Then the average (3000 points) is shown and stored as highscore using the player name "Alice & Bob"
- add background music in menu scenes
    - used the music from USDX. Added license info in about scene.
- minor bug fixes and refactoring

### Closes Issue(s)

close #305
close #279 
close #296 

### More

Will create a new release soon after merge.


basisbit:
could you maybe add a screenshot of the common score? Is that per difficulty level or over all difficulties together?

PR #317, opened 2022-07-08
Easier song folder config on Android

achimmihca:
### What does this PR do?

A small PR.
- show path that is required for song folders on Android
    - Android does not allow arbitrary access to the files system.
        - Notably txt files are only visible to Unity's C# file system API when paths are inside the app specific folder
        - The app specific folder can be found using Android's `getExternalFilesDirs`
            - I got the initial code to query this from a [blog post](http://anja-haumann.de/unity-how-to-save-on-sd-card/).
            - Don't be confused with Android's use of "internal storage" (requires root access), "primary external storage" (app specific storage on internal memory), "secondary external storage" (app specific storage on sd card)
        - fun fact: the app specific folder on the internal storage is the what Unity uses as `Application.persistentDataPath`
- added button to add the correct paths for SD card and internal storage
    - because Android mounts the storage at some paths that no normal user will be able to guess.
- changed initial songDirs of the settings
    - now set in LoadingSceneControl because `Application.persistentDataPath` must be run on main thread.
    - on Android it adds a path to both, internal and external storage
    - on other platforms, it adds a path to `Application.persistentDataPath`

Now, one can save a song library on an SD card and use it with Android.

### DIFF
--- a/UltraStar Play/Assets/Common/Model/SettingsManager.cs
+++ b/UltraStar Play/Assets/Common/Model/SettingsManager.cs
@@ -76,9 +76,6 @@ public class SettingsManager : MonoBehaviour
             {
                 UnityEngine.Debug.LogWarning($"Settings file not found. Creating default settings at {loadedSettingsPath}.");
                 settings = new Settings();
-                // Can only access Application.persistentDataPath from main thread. Thus, set the default songDirs in a coroutine.
-                StartCoroutine(CoroutineUtils.ExecuteAfterDelayInFrames(1,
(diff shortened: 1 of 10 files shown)


achimmihca:
Buttons are hidden when the song folders have already been added
Screenshot_20220708-181136_UltraStar Play

Screenshot_20220708-181152_UltraStar Play

basisbit (approved):
Lgtm. Thanks!

---
I am very much disappointed of Unity that they don't handle this in a better and more generic / platform independent way...

Issue #186, opened 2020-10-04
Improved UI


timgoeller:
I thought a bit about how the UI of Ultrastar Play could be improved. The current UI is based on the old UltraStar Deluxe interface, which feels a bit dusty by now. I think something a bit more color- and playful but clear would be nice. I made a mockup for how the main menu could look in a more modern look&feel. That kind of style could be applied to the whole program. What do you think about it? If you like it, I'll try to create a pull request that updates the application UI accordingly.

![chrome_wDz50Pewwc](https://user-images.githubusercontent.com/10633973/95021340-b2c08f80-0670-11eb-8b91-6c39590d5aa1.png)

Color palette used: https://flatuicolors.com/palette/au
Fonts used (all under Open Font License): Fredoka One, Permanent Marker

basisbit:
@timgoeller thank you for creating that draft theme! Personally, I really like that color combination and think that this should work well for touch-input-devices as well as typical desktop targets.
How would you suggest coloring the "mouse-over"/pre-select effect - should buttons just get lighter/darker or so when you are about to select them / when they have the focus?

timgoeller:
I think I'd use some kind of white border for hovering, like this (and maybe only a very slight color change on click):

![image](https://user-images.githubusercontent.com/10633973/95193713-89207900-07d4-11eb-88ec-23e52e35d2d3.png)

achimmihca:
If you want to implement this, please try to make use of prefabs as much as possible.

timgoeller:
I changed the main scene in Unity to look like the mockup. I think changing the style should be doable for the whole application. However I'm currently wondering if it might make sense to switch to the new UI Toolkit. It might be easier to switch now than having to redo everything later. What do you think?

achimmihca:
I hope so, yes.
Especially as it seems to ship with a flex-layout engine, which probably will make layout much easier.
Furthermore, there seems to be css-ability for colors etc. I think design can be much faster iterated and polished this way.

On the other hand, it is experimental in its current state. Then again, if we would migrate later then we still would have to change everything instead of just a little.
Only thing I fear are bugs that cannot be worked around, such that one is stuck with unlucky tools.
I hope they will release UI Toolkit next year. However, Unity also took super long before releasing the new input system. Better not raise expectations too high when it comes to Unity release features.

Anyway, I think one should try the current state of UI Toolkit and see if it is well usable already.
If you want, give it a shot 👍 

---
[...]

PR #322, opened 2022-08-10
Anst/all new stuff

achimmihca:
### What does this PR do?

This is a mix of fixes and new stuff.
I have this split up in branches, so could create smaller pull requests if needed.

**Fixes**
- The backslash replacement for Windows paths was not rendered in the built game. I just replaced it with a pipe.
    - This workaround will be removed anyway when Unity provides an option for [this crap](https://forum.unity.com/threads/preventing-escaped-characters-in-textfield.1071425/).
- do not play background music in RecordingOptionsScene. It messes up the mic calibration
- muted audio was reset on scene change. Now it stays muted.
- IndexOutOfRangeException in VoicesBuilder
    - I had a txt file with a trailing space for sentence end. Normally this would be `- 123 130` for example, but it was `- 123 `. Thus, splitting on space and parsing the second number was throwing an excaption
- Related to the IndexOutOfRangeException: if parsing a song fails, then it is not attempted again.
    -  In my case, the "broken" song was attempted to be parsed every frame, which was leading to a memory leak (not quite sure why)
    - Now, the song is not parsed again. A failed song cannot be started from the SongSelectScene.

**New Styles**
- Buttons use round corners, and bold text. And text color is now full white (#ffffff) anymore, but a little bit darker.
    - Small changes, but it looks much better now.
- Scrollbars are now smaller, and use all available space (had to change some built-in Unity styles)
- I changed the MainScene layout. Took inspiration from [#186](https://user-images.githubusercontent.com/10633973/95021340-b2c08f80-0670-11eb-8b91-6c39590d5aa1.png)
    - the "Sing" button is now dominant (size and shadow). I think this makes sense.
    - Motivation: With "About" button and "Credits" button, it was getting crowded. Thus, I moved them to a corner as small round buttons.

**Credits scene**
- I want to give everyone involved in the project a stage to show appreciation
    - An entry in the credits would also make a possible reward for backers when I create a Kickstarter
- To make this more engaging, the credits scene includes a mini-game: you play as a mic that collects falling letters
    - The name of each CreditEntry is "exploding", the letters fall down more or less randomly
- Some entries have golden / silver background to highlight outstanding contributions
- all images are from pixabay, publicdomainvectors, etc. See `sources.txt` for details
- the CreditEntries for contributors could probably be generated automatically from Git.
    - but for now, one can just maintain this list manually
- Fun fact: I first planned to control the mic's position by humming (reusing our pitch detection). But this did not work well.

**Scene change animation**
- I implemented this as a lib. For details see [GitHub](https://github.com/achimmihca/SceneChangeAnimations) and the [Unity forum](https://forum.unity.com/threads/scene-transition-animation.1308786/)
    - I imagine it is a usual feature for a game to transition between scenes. But to my surprise, I did not find anything ready-to-use for UIToolkit
- using the lib, I re-implemented the scene change animation from USDX. Including the good old "swoosh" sound that is played.
- Animation sound volume is configurable. Use volume 0 to turn it off.
- The scene change animation can be turned off in design settings.


achimmihca:
New main scene. Bottom right: "credits" button and "about" button:
2022-08-10 19_17_27-UltraStar Play

Credits scene with falling letters that can be collected:
2022-08-10 19_18_35-UltraStar Play

basisbit (approved):
Lgtm. Love the mini-game easteregg!

PR #326, opened 2022-10-05
Anst/fixes

### LINKED ISSUE #320: START tag is not being considered
Issue type: Bug report

### Actual behaviour
The #START tag are being ignored and the song is not jumping from the defined start timeline.

### Expected behaviour
The should start at the defined timeline on the #START tag

### Steps to reproduce

1. Open the ultrastarplay.exe
2. Select a song with a #START tag defined

achimmihca:
### What does this PR do?

Minor fixes.

- save PREVIEWSTART and PREVIEWEND tags
- consider START and END tag of txt files
- do not show lyrics on notes by default because there is an issue if there is not enough space
- show translated message when no songs found
    - the translated message informs the user that songs can be downloaded/added in the settings
- do not show "Is admin" and "enqueue songs" in companion app options because these options are not used yet
- better keyboard / gamepad navigation (FocusableNavigator)
    - custom navigation targets (if at element A and navigating in direction D then focus element B)
    - use "rectangle distance" instead of center distance
    - only consider VisualElements that are left/right/above/below the current VisualElements left side/right side/top side/bottom side
    - immedeately leave readonly TextFields because no cursor is visible. Thus, navigating the cursor would be weird.
    - do not focus VisualElements if these are disabled/unfocusable

### Closes Issue(s)

close #320


achimmihca:
Did a rebase because I changed a translation file via poeditor.com
I expected poeditor to create a PR, but it does not. Will now create a dedicated branch for new translation files.

basisbit (approved):
looks good. Thanks for the PR!

PR #331, opened 2022-10-08
Added spanish translation

BanCrash:
### What does this PR do?

Added spanish translation. NOTE: This is not tested since I never used before Unity and I can't compile the project for some reason :/

### Closes Issue(s)

None I think.

### Motivation

Adding spanish translation.

### More

### Additional Notes

As I said, this is not tested since I never used before Unity and I can't compile the project for some reason. I duplicated english properties and then translated. I readed the documentation to see how to properly translate it but since I don't have any experience I'm not sure if I did it well.

### DIFF
--- a/UltraStar Play/Packages/playshared/Runtime/Resources/Translations/messages_es.properties
+++ b/UltraStar Play/Packages/playshared/Runtime/Resources/Translations/messages_es.properties
@@ -1,0 +1,291 @@
+﻿# General
+yes=Sí
+no=No
+ok=Confirmar
+back=Atrás
+cancel=Cancelar
+exit=Salir
+close=Cerrar
+open=Abrir
+new=Nuevo
+save=Guardar
+load=Cargar
+loading=Cargando...
+quit=Salir
+start=Empezar
+options=Opciones
+add=Añadir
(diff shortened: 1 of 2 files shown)


BanCrash:
Hi @achimmihca , since I couldn't deploy I see now that there are some issues that I have to fix. What is better, to fix them ASAP or take my time to check properly them and with that avoiding opening unnecesary PR's?

The question is because I don't know which is gonna be the release calendar. I mean, if there will be releases very often which would mean that is better to have the fixes ASAP or not and so the PR could be more tested.

achimmihca:
I'd say take your time. Needed translations may change anyway (new translations needed, other may becoming obsolete), depending on the UI.

achimmihca:
BTW, I suggest to use the new integration of poeditor.com because it provides a nice interface and collaboration to edit the translation files.

BanCrash:
> I'd say take your time. Needed translations may change anyway (new translations needed, other may becoming obsolete), depending on the UI.

Perfect! 

> BTW, I suggest to use the new integration of poeditor.com because it provides a nice interface and collaboration to edit the translation files.

I'll check it, thanks!

PR #354, opened 2023-01-03
Anst/fix custom elements and refactor styles

achimmihca:
### What does this PR do?

- fix custom VisualElement implementation, notably ItemPicker
    - object hierarchy must be added in the constructor, not in Init method
    - use properties instead of simple fields because the UIBuilder maps properties automatically to the XML attributes with matching name
    - renamed field names to match XML and UI Builder conventions
        - This means dashes instead of camelCase, e.g. `minValue` becomes `min-value`

- refactor StyleSheets
    - split MainStyles.uss into multiple files
        - CommonVariables.uss => Sets variables such as `--text-color`, `--font-size`
        - UnityControlStyles.uss => Styles for Unity controls such as `Button`, `TextField`, `ScrollView`
        - CustomControlStyles.uss => Styles for custom controls such as `ItemPicker`, `ContextMenu`
        - CommonStyles.uss => Common utility styles such as font sizes, margins, rounded corners
        - SceneSpecificStyles.uss => Styles that are used only in a specific scene
    - This is a first step towards better styling implementation.
        - A next step could be to use Sass or a similar compiler to generate the USS files

- fix help dialog title
- fix NPE in ListedItemPickerControl


basisbit (approved):
lgtm, thanks for the improvement as always 🥇

PR #335, opened 2022-11-04
Jean/json themes pull request

achimmihca:
### What does this PR do?

Implement theming.
- Themes are stored in json files
- Background is animated via custom shader
- Theme styles are applied via custom logic (working around Unity "no USS at runtime" limitation)

@jean-moreno  feel free to add to this explanation

PR #357, opened 2023-01-05
Anst/json themes

achimmihca:
### What does this PR do?

Implement JSON file based themes.
Find a demo here: https://user-images.githubusercontent.com/14967033/201637601-7b8bf196-5f2c-40ac-8b36-78fbb7c42231.gif

This PR is based on #335

**Theme files**

- Default themes are stored in `StreamingAssets/Themes`
    - On Android, the StreamingAssets need to be extracted first from the apk file.
        - This is done in AndroidStreamingAssets.cs
        - The StreamingAssets are extracted to persistentDataPath
        - The last write time of the assets is used to check whether new StreamingAssets need to be extracted
- Users can add custom themes in `persistentDataPath/Themes`
- So far, theme files include
    - Button colors
    - Font colors
    - Background settings
    - Scene transition animation time
        - The animation time is a string that includes the unit, e.g. "0.5s" or "500ms"

**Theme data structure**

- ThemeMeta is the main data structure of a theme, similar to SongMeta
    - holds the path to the theme file
    - holds the ThemeJson data structure, which corresponds to the content of the theme file
- ThemeJson is created using JsonConverter, i.e., using reflection
- I added a custom Color32Converter for the JsonConverter
    - This custom color serializer is now also used for the settings, e.g. mic profile colors
    - It serializes to hexadecimal, e.g. `new Color32(255, 0, 0)` is serialized as `"#FF0000"`
    - It deserializes hexadecimal, plus the previously used object notation
        - It can handle `{"r":255, "g":0, "b":0}` to be compatible with previous version of settings

**Background shader**

- The Background is rendered and animated via custom shader
- The UI is rendered to a RenderTexture, which is then blend with the background shader.
   - This makes the scene transition animation smoother

**ThemeManager**

- New ThemeManager is taking care of loading and applying the themes
- A ParticleSystem adds some images to the background (stars/hearts/pumpkins/
- Images (e.g. particles, song rating images) are loaded using the existing ImageManager that takes care of caching and cleanup

### Motivation

- Eye candy.
- Also it might be worth to consider user defined themes in future design.

### More

- Should we create a dedicated repo for user defined themes?
    - It would allow to download themes in-game, similar to downloading songs


basisbit (approved):
Cool

basisbit (dismissed):

basisbit:
Not sure if a user defined themes repo this "early" in the project would be a good idea. I expect them to still break regularly if stuff gets changed in the games code. Maybe after everything got a bit more established / settled in?

achimmihca:
I agree, things are likely to change again.
At least when Unity's UI Toolkit allows loading style sheets at runtime, some fields will become obsolete.

OK so I won't create a themes repo now.

Issue #342, opened 2022-11-24
Webcam support


achimmihca:
### Actual behaviour

At the moment, UltraStar Play does not support webcams.

### Expected behaviour
It would be nice to see the singer on the screen, also for recording purpose.
And for taking profile pictures (#240).

Other similar games have this feature:
- UltraStar Worldparty provides a rather basic support for integrated devices.
- From the [Vocaluxe website](https://www.vocaluxe.org/): "Singing is even more fun when you can see yourself - Vocaluxe supports webcams. You can also use it to take pictures for the player profiles"

### Implementation
Unity has an API to access the webcam. Overall it should be doable.
- https://docs.unity3d.com/ScriptReference/WebCamDevice.html
- https://docs.unity3d.com/ScriptReference/WebCamTexture.html

This repo demonstrates how to use the webcam: https://github.com/eranws/unity-webcam

PR #358, opened 2023-01-05
Basic webcam support

supernevi:
### What does this PR do?

You can configure a webcam in the new options menu:
Screenshot_20230105_175633~2

Here you can select your device with a little preview:

Screenshot_20230105_175754

If you have only one device you don't really need this step to configure a webcam, because the only one will be chosen by default.

However, the Singing scene now have a little toggle icon in the top left corner:

Screenshot_20230105_175900~2

Here you can toggle your webcam on/off:

Screenshot_20230105_175908

Tested on Windows 10 and Android

### Closes Issue(s)
#342 

### Motivation

I'm new to unity (and actually to the GitHub flow) and I want to learn it. I did a tutorial and now want to bring ultrastar a little step further :)

If that will be merged I would work on configuring the resolution and fps for a selected webcam device.
Or on taking screenshots via webcam for a player profile.

### More

Before merge please test for example the new options menu by clicking on it (your webcam should appear) and click back to options menu. Then again on webcam and back. Try it at least 5 times. On my Windows 10 machine the App (and even then the unity editor crashes). But on my Android phone everything is fine. I found in the web that it could caused of my virtual webcam from the OBS tool. 


supernevi:
Eventually I found an indication about my app crash:

image

That "RADAR_PRE_LEAK" let me found this:

https://answers.microsoft.com/en-us/windows/forum/all/radarpreleak64/a534e186-65ab-42c9-97e2-b654a713f0db

So it could be a memory problem. Since i thought C#'s Garbage-Collector handles it, I still have no clue where the problem could be... Maybe someone else have an idea?

UltraStar Play/Assets/Scenes/Options/WebcamOptions/WebcamOptionsControl.cs
```diff
@@ -0,0 +1,106 @@
...
+                }
+                webcamTexture = new WebCamTexture(device.name);
+                Log.Logger.Information("Setting current webcam to '{webcamname}'", webcamTexture.deviceName);
+                webcamTexture.Play();
+
+                webcamRenderContainer.image = webcamTexture;
+            });
+    }
+
+    private void navigateBack()
```
achimmihca, line 84: NavigateBack

UltraStar Play/Assets/Scenes/Sing/SingSceneControl.cs
```diff
@@ -98,9 +99,23 @@ public static SingSceneControl Instance
     [Inject(UxmlName = R.UxmlNames.inputLegend)]
     private VisualElement inputLegend;
 
+    [Inject(UxmlName = R.UxmlNames.webcamRenderContainer)]
```
achimmihca, line 102: Please create a new SingSceneWebcamControl for all this. I think it makes sense to create a new GameObject in the scene for this.

UltraStar Play/Assets/Common/Model/Setting/Settings.cs
```diff
@@ -12,6 +12,8 @@ public class Settings : ISettings
     public List<PlayerProfile> PlayerProfiles { get; set; } = CreateDefaultPlayerProfiles();
     public List<MicProfile> MicProfiles { get; set; } = new();
     public string LastMicProfileNameInRecordingOptionsScene { get; set; }
+    public string LastWebcamNameInWebcamOptionsScene { get; set; }
+    public bool LastGameWebcamActivated { get; set; }
```
achimmihca, line 16: I suggest to add `UseWebcamAsBackgroundInSingScene` or similar

achimmihca (commented):
Quite nice so far! Some general thoughts:
- Please don't show the webcam button all the time in the SingScene.
    - Instead, I would expect a checkbox in the WebcamOptionsScene "Use webcam as background when singing"
    - Or add a "toggle webcam" button to the context menu of the SingScene
- You may want to show a "No webcam found" message when no webcam has been found.
    - Does the app require a restart to find a new webcam? If yes, then this should be communicated to the user in the "no webcam found" message.
- I think your doll is pretty creepy ;)

I have no idea where users expect the webcam image in the sing scene. Covering the whole background seems like a lot to me. How do other singing games make use of the webcam (Vocaluxe, Just Sing, We Sing, USDX, etc.)?

Thanks for the contribution!

achimmihca:
About the memory leak: the master branch uses an alpha version of Unity (because it fixes other issues).
I could imagine that the Unity LTS version does not have this issue.

So maybe try it with Unity 2021.3 or 2022.2 to check if it is a bug in Unity

achimmihca:
I think it makes sense to associate the image with the player profile, i.e., at most one webcam image per player profile.

- The player profile image chooser needs a new entry for "webcam image"
    - When the webcam image should be used then a new button is needed that
        - (a) if there is no saved webcam image yet then save a snapshot
        - (b) otherwise delete the currently saved webcam image to start over
- One could save the image file using the player profile index.
    - I'd suggest the location `Application.persistentDataPath + "/Webcam/ProfileImages"`
- If such a file exists, then this is used as webcam texture for this player profile.

supernevi:
> Quite nice so far! Some general thoughts:
> 
> * Please don't show the webcam button all the time in the SingScene.
>   
>   * Instead, I would expect a checkbox in the WebcamOptionsScene "Use webcam as background when singing"
>   * Or add a "toggle webcam" button to the context menu of the SingScene
Aye! I think both should be useful

> * You may want to show a "No webcam found" message when no webcam has been found.
>   
>   * Does the app require a restart to find a new webcam? If yes, then this should be communicated to the user in the "no webcam found" message.
`WebCamTexture.devices` returns always the current connected webcams. If i disconnect my webcam in-game i just have to reopen the webcam options
> * I think your doll is pretty creepy ;)
Sorry xD
> I have no idea where users expect the webcam image in the sing scene. Covering the whole background seems like a lot to me. How do other singing games make use of the webcam (Vocaluxe, Just Sing, We Sing, USDX, etc.)?
I just wanted to be inspired from Vocaluxe (v0.4.1) and USDX (UltraStar Deluxe Worldparty 12.07), but in both there doesn't appear a webcam in the singing screen, like it isn't fully implemented. So I decided by myself and show it in fullscreen like a video. I think it is most useful. And if we have other ideas, we could do it configurable maybe (or just change it, after some feedback)
> 
> Thanks for the contribution!
You'r welcome, i try my best :)

supernevi:
> About the memory leak: the master branch uses an alpha version of Unity (because it fixes other issues). I could imagine that the Unity LTS version does not have this issue.
> 
> So maybe try it with Unity 2021.3 or 2022.2 to check if it is a bug in Unity

Yes it seems to be a problem of the unity editor version. I just tried it with 2022.2.1f1 and all works perfectly all right. I'm going to develop with this version now

achimmihca:
> if we have other ideas, we could do it configurable maybe (or just change it, after some feedback)

Works for me.

BTW: No idea why the tests failed. Seems to be a license issue, but on my other PR it was working. Let's try again with you next push...

achimmihca:
> I just tried it with 2022.2.1f1 and all works perfectly fine.

Well, I will downgrade to this version. Unity 2023 is just not ready yet.

UltraStar Play/Assets/Scenes/Sing/SingSceneControl.cs
supernevi: Unfortunately I do not understand excatly what you mean... Can you specify it a bit more for me?

supernevi:
Btw: Current master seems to be broken. Can't build locally anymore

supernevi:
You:

> Please create a new SingSceneWebcamControl for all this. I think it makes sense to create a new GameObject in the scene for this.

Me:
> Unfortunately I do not understand excatly what you mean... Can you specify it a bit more for me?

This is the only Task left I think

achimmihca:
> Unfortunately I do not understand excatly what you mean... Can you specify it a bit more for me?

Move the code for handling the webcam in the SingScene to a new class `SingSceneWebcamControl`.
A new GameObject in the scene hierarchy can be created to host this script instance (see screenshot).

sing-scene-hierarchy

achimmihca:
Optionally, you can add a binding for the SingSceneWebcamControl to make it available everywhere during injection (i.e. `[Inject] private SingSceneWebcamControl webcamControl` will work in other scripts).

Take a look at `SingSceneControl.GetBindings` method, for example the SongAudioPlayer.

supernevi:
OK thanks. Will do it on Tuesday I think

achimmihca:
@supernevi , is your branch behind master? At least the test of StandaloneLinux64 should work (Android platform has an issue because JDK is not found somehow).

UltraStar Play/Assets/Scenes/Sing/SingSceneWebcamControl.cs
```diff
@@ -0,0 +1,83 @@
...
+
+public class SingSceneWebcamControl : MonoBehaviour, INeedInjection
+{
+    [Inject]
+    private Settings settings;
+
+    [Inject(UxmlName = R.UxmlNames.webcamRenderContainer)]
+    private Image webcamRenderContainer;
+
+    private WebCamTexture webcampTexture;
```
achimmihca, line 22: typo `webcamTexture`

achimmihca:
otherwise looks good to me

Issue #307, opened 2022-05-31
UI input not working when running main game on Android via Samsung DeX, Pico 4 VR, Nintendo Switch


achimmihca:
### Actual behaviour
I connected my phone to a TV via USB-C to HDMI-adaptor.
I also connected a bluetooth keyboard and mouse to the adaptor.

The keyboard works as expected. And the mouse moves as expected outside of the game.
But the in-game buttons of UltraStar Play do not react to mouse clicks.

This is weird because the same bluetooth mouse works when runnig the main game not via DeX but directly on Android.
Needs investigation. Seems to be an issue with DeX.

Other people can confirm?

### Expected behaviour
The bluetooth mouse should work with Samsung DeX.

### Steps to reproduce
1. Connect Android device to external monitor (in my case Galaxy S10e)
2. Connect bluetooth mouse with the Android device
3. Try to use do mouse clicks in UltraStar Play

### Details
- UltraStar Play version: 0.5
- Operating system: Android 12

achimmihca:
> But the in-game buttons of UltraStar Play do not react to mouse clicks.

Actually, the very first click sets the focus to a button. But no other action is performed. And afterward, no other button gets the focus when clicked.
Could also be a Unity bug.

achimmihca:
> Could also be a Unity bug.

I created a bug report.
See also the thread on the Unity forum: https://forum.unity.com/threads/mouse-clicks-not-working-in-samsung-dex-desktop-experience.1335983/

achimmihca:
> I created a bug report.

Unity confirmed the bug, although it changed a little in newer Unity versions (clicks work in general but everything is damn slow after few seconds). Anyway, looks like this is beeing worked on by Unity.

achimmihca:
The issue only occurs when there is an EventSystem in the scene. The EventSystem is used for the custom InputActions of UltraStar Play.

The Companion App is not affected because it does not require the custom InputActions and hence no EventSystem component in the scene.

achimmihca:
Users found that this issue also occurs on the Pico 4 VR headset. Again, the issue only occurrs when an EventSystem is present in the scene.
Others users found that similar issue occurs on Nintendo Switch when running custom Android on it.

All these issues seem to have the same origin, it seems like UI Toolkit is missing some events.

achimmihca:
I created a minimal example to test this: https://github.com/achimmihca/UnityButtonClickIssueWithSamsungDeX

[...]

PR #367, opened 2023-01-14
Anst/fixes

achimmihca:
### What does this PR do?

Minor fixes.
- do not apply apply custom theme in song editor
- show general error message with log file location if an error occurred in the loading scene
- opt-in for custom event system on Android
    - This is a workaround for #307
    - If the option is not set, then the EventSystem is disabled (only on Android).
    - The option can be set in the development options

### Closes Issue(s)

workaround for #307


basisbit (approved):

supernevi:
No Click-Event-Issues anymore :)

Here a pic that you can imagine how it looks like. 
Unfortunately somehow it is not possible to download the demo songs.
But after manualy putting songs into the folder (and restart the game) it "works". you can sing but loading the image and the video failes. here a short video that you can imagine what it looks like in a meta quest 2 :)

https://user-images.githubusercontent.com/22367297/212622845-05916403-97f0-49d0-bbf6-a061f3516480.mp4

btw: only 20 FPS :/

achimmihca:
> btw: only 20 FPS :/

There are probably some performance optimiztations that could be done.
But it's not a prio at the moment.

PR #369, opened 2023-01-16
Anst/usability fixes and song editor improvements

achimmihca:
### What does this PR do?

Minor improvements with respect to usability and some fixes, notably for Android song folders and the song editor.

- button to open wiki in help dialog
    - such a button is also shown when no song folders have been found
    - and when an error occurred in the LoadingScene
- new help dialog in ContentDownloadScene
- try-catch around initial settings creation
    - Paths are accessed to set default locations. I could imagine that this may cause issues on some devices, which could explain why some users only see a black screen after opening the app.
- show song folder path on Android as "internal", "sd card", or "other"
    - the cryptic paths of Android seem to confuse people. I think it is better to hide this complexity.
    - the paths "internal", "sd card", and "other" can be toggled easily.
    - this makes the buttons obsolete for "add internal song folder" and "add sd card song folder". Thus, these buttons have been removed.
- handle song file without any notes

**Song editor fixes and improvements**
- layers (i.e. their notes) can be made read-only
    - the copy&paste layer is now read-only
- unify handling of notes song vs. notes in layers
    - There is now an differentiation between `EnumLayer` and a `VoiceLayer`
- chaning playback position with the cursor in the lyrics area is now optional
- the lyrics of a single sentence can be edited
    - either use the context menu or double click a sentence
- notes are split when editing a single note and using a semicolon (end of syllable) or space (end of word)
- hide lyrics on notes when zoomed out too much
- button to open song folder
- bunch of other fixes

Besides these changes
- removed code for Kickstarter dialog

### Motivation

- The links to wiki pages and hiding the cryptic part of the path on Android has been inspired by feedback on Discord and on Google Play for the pre-release.
- The song editor has been fixed an improved because other people seem to start using it :)

### Additional Notes

- I'd like to create a new pre-release soon with these fixes and improvements


supernevi:
> * Paths are accessed to set default locations. I could imagine that this may cause issues on some devices, which could explain why some users only see a black screen after opening the app.

unfortunately the black screen still exists if i try to build a vr app

UltraStar Play/Assets/Common/Model/SettingsManager.cs
```diff
@@ -93,17 +94,54 @@ private Settings CreateDefaultSettings()
...
+        }
+
         return defaultSettings;
     }
 
+    private MicProfile CreateDefaultMicProfile()
+    {
+        if (Microphone.devices.Length <= 0)
+        {
+            return null;
```
basisbit, line 137: Might be worth it to add some logging here, or maybe where ever the CreateDefaultSettings() is called.

basisbit (approved):
lgtm. Looking forward to the upcoming Android release, as well as the first steam release 🥳

PR #428, opened 2023-12-27
Change saved song file encoding to UTF-8 without BOM

Tuupertunut:
### What does this PR do?

Currently Ultrastar Play song editor saves all song files as UTF-8 with BOM. This happens because in code the encoding is `System.Text.Encoding.UTF8`, which produces a BOM according to [documentation](https://learn.microsoft.com/en-us/dotnet/api/system.text.utf8encoding?view=net-8.0#remarks).

This PR changes the encoding for saved song files to UTF-8 without BOM.

### Motivation

The [Ultrastar format specification](https://usdx.eu/format/) was [recently updated](https://github.com/UltraStar-Deluxe/format/pull/45). It now specifies that all song files should use UTF-8 without BOM as their encoding. There are also tools in the Ultrastar ecosystem, such as Performous Composer, that fail to open song files with BOM.

There is some discussion about UTF-8 BOMs in this Ultrastar format issue https://github.com/UltraStar-Deluxe/format/issues/42

### Additional Notes

I have not tested this code. The change seems trivial enough that I'm confident it would work.

### DIFF
--- a/UltraStar Play/Packages/playshared/Runtime/Song/UltraStarSongFileWriter.cs
+++ b/UltraStar Play/Packages/playshared/Runtime/Song/UltraStarSongFileWriter.cs
@@ -9,7 +9,8 @@ public static class UltraStarSongFileWriter
     public static void WriteFile(string absolutePath, SongMeta songMeta)
     {
         string txtFileContent = GetTxtFileContent(songMeta);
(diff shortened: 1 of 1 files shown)


basisbit (approved):
lgtm, thanks for the improvement!

PR #429, opened 2023-12-29
Fix build failing in dependency setup

Tuupertunut:
### What does this PR do?

This PR fixes the currently failing build. It fixes a bug in the dependency download scripts that caused libraries to not be found. See commit message for technical details.

### Motivation

The build was failing. There was a time bomb like behavior in the dependency download scripts where builds could break at any time if a library developer decided to make more than 100 commits after the one that Ultrastar Play depended on. The SharpZipLib dependency failed to download because of this bug.


achimmihca:
The download dependencies bash scripts are not a very good solution in the first place.
I would love to have a better tool, something like Maven, Gradle or Grunt but for Unity.

Maybe [Cake (C# Make)](https://cakebuild.net/extensions/cake-unity3d/) would be a better solution.

I think writing the download dependencies scripts in C# would make sense given that Unity also uses C#

Another approach could be to add a Menu Item in the Unity editor that runs the corresponding setup code in C#.
However, this has the drawback that the Unity editor must open with a broken setup first. Not sure whether this will work well.

basisbit (approved):

Issue #435, opened 2024-06-26
Reset test context between Unity Play Mode tests


achimmihca:
### Actual behaviour
When running all Play Mode tests in sequence, then there are test failures. But when executed one after another, then they finish successfully.

This is because Unity does not reset the whole application context between Play Mode tests, which seems to be a known shortcoming of Unity (see [forum thread](https://forum.unity.com/threads/running-tests-consecutively.781847/).

This leads to a failing CI pipeline, although the game logic is correct. The test logic is broken.

### Expected behaviour
Executing a test alone or in multiple sequence should not make a difference.

### Implementation Hints
One has to reset all application state manually in a TearDown method, notably DontDestroyOnLoad GameObjects, and static variables.

achimmihca:
Maybe there is some better lib to handle tests in Unity.
Currently, Responsible is used, but I feel like it is not adding much value.

Besides the shortcoming of Unity, I am also not happy with the current test context in general.
Maybe one finds a better solution design that uses existing Dependency Injection to wire test instances instead of production instances.

PR #469, opened 2024-09-02
Removed outdated packages from manifest.json

qamil95:
### What does this PR do?

Removes outdated packages from package.json, just to try what would happen.

### Closes Issue(s)

Nothing, only trying to check why build is failing.

### Motivation

No possibility to download UltraStar Play after merging melody mania changes.

### More

### Additional Notes

Could you approve starting build? I don't have unity license so I can't run it on my forked branch, I also don't know if it would resolve the issue but why not check?


achimmihca:
I don't understand what you try to achieve with this PR.

At least `com.unity.render-pipelines.universal` is important to build the project. This is the rendering pipeline used for the Unity project.

> No possibility to download UltraStar Play after merging melody mania changes.

You can still download v0.9.0 of UltraStar Play.

Further, you can create a more recent build by yourself that includes the merged code from Melody Mania.
Therefore, follow the instructions in the wiki: https://github.com/UltraStar-Deluxe/Play/wiki/Download-and-Install#download-and-build-from-source-code 

Only reason that there is no updated, ready to use build yet is that nobody took the time to create it.
UltraStar Play is welcoming maintainers of the build (e.g. DevOps stuff).

Anyway, no changes to the package.json should be needed.

achimmihca:
> I don't have unity license so I can't run it on my forked branch

There is a free license available that should get you started.

> I also don't know if it would resolve the issue but why not check

Not sure what you mean. But if you mean the failed pipeline, then this is mostly due to failing tests because of https://github.com/UltraStar-Deluxe/Play/issues/435
One short-term workaround might be to disable the tests. But it is not a graceful solution.

AnsisMalins:
I'm closing this pull request because it is superseded by my pull requests that remove even more packages. You have the right idea though, [qamil95](https://github.com/qamil95). Re: Universal Render Pipeline: it can be removed, but then its configuration assets need to be removed too, else they will be left dangling.

PR #485, opened 2024-10-17
feat: Batch Isolate Vocals should filter out songs with unsupported a…

corykim:
…udio formats

### What does this PR do?

This PR modifies the Batch Isolate Vocals mod by filtering out any songs with unsupported audio formats. We show a disabled toggle for any song that is not supported.

### Closes Issue(s)

N/A

### Motivation

The reason we do this is because the audio separation job cancels if it encounters an unsupported song, and the entire batch gets canceled. 

### More

N/A

### Additional Notes

N/A

### DIFF
--- a/UltraStar Play/Assets/StreamingAssets/Mods/BatchIsolateVocals/1-BatchIsolateVocalsModSettings.cs
+++ b/UltraStar Play/Assets/StreamingAssets/Mods/BatchIsolateVocals/1-BatchIsolateVocalsModSettings.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.IO;
 using System.Linq;
 using UniInject;
 using UniRx;
@@ -39,7 +40,7 @@ public class BatchIsolateVocalsModSettings : IModSettings
 
     private void OnSelectAll()
     {
-        toggles.ForEach(toggle => toggle.value = true);
+        toggles.FindAll(toggle => toggle.enabledSelf).ForEach(toggle => toggle.value = true);
     }
 
     private void OnDeselectAll()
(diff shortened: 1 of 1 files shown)


UltraStar Play/Assets/StreamingAssets/Mods/BatchIsolateVocals/1-BatchIsolateVocalsModSettings.cs
```diff
@@ -152,13 +153,33 @@ public VisualElement CreateVisualElement()
...
+                    toggle.SetEnabled(false);
+                }
+                toggles.Add(toggle);
                 toggleContainer.Add(toggle);
             }
 
             return toggleContainer;
         }
+
+        private readonly string[] supportedExtensions = { "wav", "mp3", "ogg", "m4a", "wma", "flac" };
```
basisbit, line 168: file extensions tend to not be a reliable indicator of what is in them, especially for media files. Also, supported file types and file formats depends on platform where the game is running on.
How about instead of filtering based on file type, maybe handle whatever error may happen during scanning, by adding some try catch?

basisbit (changes requested):

UltraStar Play/Assets/StreamingAssets/Mods/BatchIsolateVocals/1-BatchIsolateVocalsModSettings.cs
achimmihca: Current implementation for Vocals Isolation uses [SpleeterMsvcExe](https://github.com/wudicgi/SpleeterMsvcExe), which uses ffmpeg. I am unsure what file types are supported there.
Anyway, the supported file types for Vocals Isolation may differ from what the game supports for playback.

The approach with file extensions is probably "good enough" for most cases. I would be fine with that.
However, the better solution would be more forgiving error handling. But it requires adjustments of the current job system. I can take a look at this soon.

AnsisMalins: This change does not make anything worse, does it? What is our standard of quality for contributions?

achimmihca: As I said, I am fine with this. It is only for a mod anyway.

achimmihca:
Will include this change in next version of Melody Mania.

PR #486, opened 2024-10-23
Remove Sqlite

AnsisMalins:
### What does this PR do?

Remove code related to Sqlite.

### Closes Issue(s)

None.

### Motivation

The code seems to be unused.

### More

None.

### Additional Notes

None.


achimmihca:
Motivation for SQLite was compatibility with score database of UltraStar Deluxe (also Vocaluxe, UltraStar World Party).

Melody Mania uses it for a mod. But it was more a proof of concept.

But I am fine with removing this. It is "just a game" after all, losing old scores is not a big deal.

AnsisMalins:
Ah, I see. I missed that because neither Unity nor Visual Studio will tell me about things used by mods because they're just loose .cs files.

achimmihca (approved):

achimmihca (approved):

achimmihca:
@AnsisMalins this is a Draft Pull Request. Is this intended?

AnsisMalins:
Yes. A mod depends on Sqlite so merging this now would break it. At the same time, I would like to investigate if the mod has to depend on Sqlite, and what to do about mods overall.

My thoughts about mods:
* If some piece of the game is unused except by a mod, should it then not be part of that mod?
* But then if multiple mods want the same thing, we now need a system of dependencies for mods, and before you know it, we have implemented another npm/rpm/deb/nuget/whatever, poorly.
* But this is an open source project, why not promote all the bundled mods into proper code, and tell people to commit to UltraStar Play instead of writing mods. Things not everyone might want can then be behind feature switches. The game's UI need not change at all.

AnsisMalins:
But then also what about piracy mods.

achimmihca:
> If some piece of the game is unused except by a mod, should it then not be part of that mod?

Yes.

No worries, I should be able to include the DLL of SQLite, plus utility code in the mod.

> But then if multiple mods want the same thing, we now need a system of dependencies for mods, and before you know it, we have implemented another npm/rpm/deb/nuget/whatever, poorly.

This is not intended.

> But this is an open source project, why not promote all the bundled mods into proper code, and tell people to commit to UltraStar Play instead of writing mods.

I agree, mostly.
- Overall, I am fine with removing the current mod system from UltraStar Play.
  - Stuff that is an improvement for the game can become part of the game.
  - Furthermore, the current mod system requires runtime loading of C#, which does not work with IL2CPP backend (won't work on iOS, Android built for Google Play)

That said, motivation for the current mod system was
- enable users to extend Melody Mania independently, even if I personally cannot maintain their feature wishes
 - note that not all of MM is open source due to license issues
 - note that UltraStar Play only has few maintainers
- enable users to integrate stuff they want (e.g. some song service), even if I personally would not be able to do so (e.g. because of limited resources or legal issues)
- make it easier to contribute and try out new game ideas (e.g. game modifiers), even if these should not find their way into UltraStar Play.

> But then also what about piracy mods

Not sure what you mean with that.
AFAIK, we should not be reliable for what other people do with these tools.

---
In conclusion,
- I'd say you are free to kick out SQLite from UltraStar Play
- I am fine with removing the current mod system from UltraStar Play

achimmihca:
Will close this for now.
Not a big difference anyway.

SQLite is still included for now because it is used in the mod to make use of USDX databases.

PR #490, opened 2024-10-23
Remove ffmpeg

AnsisMalins:
### What does this PR do?

Removes dead code associated with ffmpeg. Ffmpeg is currently not integrated because the integration is proprietary. If/when we add ffmpeg in an open source compatible way, this code can be pulled out of the history if need be.

### Closes Issue(s)

None.

### Motivation

Tidy.

### More

None.

### Additional Notes

None.


achimmihca:
I am fine with removing this.
This was also on my agenda because Melody Mania is using libVLC by now.

basisbit (approved):

PR #491, opened 2024-10-23
Remove LibVLC

AnsisMalins:
### What does this PR do?

Removes dead code associated with LibVLC. LibVLC is currently not integrated because the integration is proprietary. If/when we add LibVLC in an open source compatible way, this code can be pulled out of the history if need be.

### Closes Issue(s)

None.

### Motivation

Tidy.

### More

None.

### Additional Notes

None.


achimmihca:
I am fine with removing this.
Melody Mania will continue to use libVLC, but I see why it will not work for UltraStar Play, at least as it is now.

basisbit (approved):

basisbit:
Imho we should add libvlc to USPlay. Just don't use that non-free unity marketplace thing. I'll merge this MR for now to make it easier for a new fresh attempt to integrate libvlc.

PR #496, opened 2024-10-25
Remove ThreadPool folder

AnsisMalins:
### What does this PR do?

Removes ThreadPool folder

### Closes Issue(s)

None.

### Motivation

It contains just a README file and nothing else.

### More

None.

### Additional Notes

Unity has its own ThreadPool implementation now.


achimmihca:
Yes, should be removed.
I think there once was some third-party ThreadPool implementation. But I removed it as it was not needed in the first place.

basisbit (approved):

PR #500, opened 2024-11-05
Change SongSearch behaviour

Nottrex:
### What does this PR do?

Change SongSearch to search based on all individual words in the SearchText instead of the literal SearchString

### Motivation

The SongSearch currently searches for the whole string as given. This causes problems and confusions when using to many whitespaces ("ABBA " or "Imagine  Dragons" with two spaces would not find anything) and it makes it difficult to search for artist and title at the same time ("Queen Bohemian" does not find anything). I have seen a lot of people confused about both of these problems.

The more intuitive behaviour would be to split the search string at all whitespaces and to search for each part individually. This small edit should create this behaviour.

### Additional Notes

One could also add a small button to switch between the two search variants, but I think the proposed variant would almost always be preferred.
 (I did not test the changes because I did not want to set up everything for this tiny commit. So I don't know if StringSplitOptions needs to be imported or something)

### DIFF
--- a/UltraStar Play/Assets/Scenes/SongSelect/Search/SongSearchControl.cs
+++ b/UltraStar Play/Assets/Scenes/SongSelect/Search/SongSearchControl.cs
@@ -262,11 +262,15 @@ public class SongSearchControl : INeedInjection, IInjectionFinishedListener
         string searchText = GetRawSearchText() != "#"
             ? GetSearchText()
             : "";
-        string[] searchTexts = searchText.Split(new char[0], StringSplitOptions.RemoveEmptyEntries);
-        List<SongMeta> filteredSongs = songMetas;
-        for (string searchWord in searchTexts) {
-            filteredSongs = filteredSongs.Where(songMeta => SongMetaMatchesSearchedProperties(songMeta, searchWord)).ToList();
+        if (searchText.IsNullOrEmpty()) {
+            return songMetas;
         }
+        // Split searchText at whitespaces and match each word individually
+        string[] searchTexts = searchText.Split(new char[0], StringSplitOptions.RemoveEmptyEntries);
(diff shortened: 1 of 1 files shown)


UltraStar Play/Assets/Scenes/SongSelect/Search/SongSearchControl.cs
```diff
@@ -262,10 +262,11 @@ public List<SongMeta> GetFilteredSongMetas(List<SongMeta> songMetas)
         string searchText = GetRawSearchText() != "#"
             ? GetSearchText()
             : "";
-        List<SongMeta> filteredSongs = songMetas
-            .Where(songMeta => searchText.IsNullOrEmpty()
-                               || SongMetaMatchesSearchedProperties(songMeta, searchText))
-            .ToList();
+        string[] searchTexts = searchText.Split(new char[0], StringSplitOptions.RemoveEmptyEntries);
+        List<SongMeta> filteredSongs = songMetas;
+        for (string searchWord in searchTexts) {
```
achimmihca, line 267: Thanks for the PR, I think it is a neat idea.

One suggestion: I would prefer a single Where-clause over this imperative filtering in a loop.

For example, consider something like this (not tested):

```
List<SongMeta> filteredSongs = songMetas
            .Where(songMeta => searchTexts.IsNullOrEmpty()
                               || searchTexts.AllMatch(searchText => SongMetaMatchesSearchedProperties(songMeta, searchText)))
            .ToList();
```

IMO this would make it more obvious that a song must match all search texts.

achimmihca (changes requested):

Nottrex:
Yes this looks much nicer. I changed the code.

achimmihca (approved):

achimmihca:
Looks good to me, thanks

Issue #526, opened 2025-05-30
Play Mode tests in CI


achimmihca:
### Actual behaviour
Play Mode tests of the main game are currently not executed for a Pull Request. Only Edit Mode tests are executed.

Reason is that there are some issues that occur only in the CI, but not on Windows.

### Expected behaviour
Play Mode tests should run and pass as part of a Pull Request.

### Implementation hints
Needs a developer working on Linux to fix this issue.
When it works on Linux in the Unity editor, it will likely also work in the CI

PR #536, opened 2025-06-01
Fix Play Mode tests on Linux

Eisbaer1832:
### What does this PR do?
This PR should fix the Play Mode tests failing on Linux.

### Closes Issue(s)
#526 

### Additional Notes
I cannot verify at the moment that automated tests via Github actions will work.
Executing `./build.sh TestMainGame` locally is confirmed to work.


achimmihca:
Thanks for tackling this issue.

One remark: 
- the tests of a Pull Request use GitHub Actions to run [GameCI](https://game.ci/) Docker images
  - see the workflow files in .github folder
  - Nuke build is only executed in the GitHub Actions workflow to resolve dependencies, but not to start Unity
- so, fixing the Nuke build is good, but might not be enough to fix the CI pipeline

achimmihca:
The pipeline currently fails because the Unity license could not be activated.
The reason is probably that the environment variables of the original repo are not passed to your PR for security reasons.

I think it should work if you define the following in your fork's settings (under `Settings > Secrets and variables > Actions`
- UNITY_EMAIL
- UNITY_PASSWORD

Your Unity credentials should then be used to aquire a free Unity license for the CI run.

Issue #569, opened 2026-02-02
Settings, Song Queue and Modifiers windows open and displaced in Song Select Menu


Kaurli:
Issue type: Bug report

### Actual behaviour
Whenever the Song Selection Menu gets opened the Song Queue and Modifiers Windows are above each other on the left side of the screen and the Settings Menu next to the two.
The buttons close, apply and "X"  buttons of these windows do not work. The other buttons of these windows seem to work.
By clicking on the Buttons of these Menus in the top right the first time, the position of the respective Window will change to the right side. The position seems the intended one. By clicking the Button an additional time it will close the window and work as intended until the Song Select Menu is closed and opened again. 

### Expected behaviour
I think these windows should be hidden upon entering the menu. Im a fresh user and didnt find pictures of this Menu so i dont know exactly.

### Steps to reproduce
- Reproducible (mark with x): [x] always, [ ] sometimes

Please provide detailed step-by-step instructions to help us understand and replicate the issue you encountered.
Try to use as few steps as possible after starting the game. Clear and concise instructions will make sure that the issue can be addressed promptly.

1. open Song Select Menu. Either at the end of a song or through the Main Menu
2. By clicking on the buttons for these mentioned windows they change the position to the right side and will be closed upon second click on the button

### Details
Please provide additional information.

- Game version: release = 1.5.1-beta6
commit_hash = d32b877e
unity_version = 6000.1.3f1
build_timestamp = 2505262013
website_link = https://ultrastar-play.com/get

- Operating System and version:  Bazzite 43 Desktop (linux)
- Song details, if applicable (e.g. audio and video file format): 
- Attach screenshots and screen recordings
- Attach a log file. To find the log file, see https://melodymania.org/how-to-report-issues

### Implementation hints
Information on how the expected behavior could be achieved (optional).

<img width="3840" height="2160" alt="Image" src="https://github.com/user-attachments/assets/67aafd6a-d4b1-4756-bc9f-7b4b0e17bf2f" />

<img width="1056" height="615" alt="Image" src="https://github.com/user-attachments/assets/aa8a3cf8-ecd7-41db-ab5d-0d8dd6b4fe49" />

Kaurli:
I couldnt attach the .log File in the Issue nor in this comment. 
Therefore I uploaded it here:
https://limewire.com/d/CJKqG#vwyYaWyxD3
If I should upload it elsewhere or the link expired let me know.

achimmihca:
[...]

Issue #538, opened 2025-06-01
Song icons can't be displayed


Eisbaer1832:
Issue type: Bug report

### Actual behaviour
When opening the song selection screen, the songs use the default ultrastar play logo as their icons.

### Expected behaviour
The songs should use the provided image files instead.

### Steps to reproduce
- Reproducible : [x] always, [ ] sometimes
- open the song selection 

I can only confirm this being a problem in the Linux builds, this may not occur on Windows.

### Details
- Game version: 1.5.1 - beta 6
- Operating System and version: Linux 64bit
- Song details: Song works in USDX

### Implementation hints
```
NullReferenceException: Object reference not set to an instance of an object
SongSelectEntryControl.UpdateSongIcons (SongSelectSongEntry songEntry) (at Assets/Scenes/SongSelect/SongList/SongSelectEntryControl.cs:528)
```

achimmihca:
Are you using UltraStar Play or Melody Mania?

From the line number mentioned in the log, it looks like playlistManager.FavoritePlaylist is null.
Does it work to mark songs as favorite?

You should find the file in `DATA_FOLDER/Playlists/Favorites.upl`, where `DATA_FOLDER` can be opened from the Development Options.

Eisbaer1832:
I'm using Ultrastar Play. Favorites.upl exists and the game apparently even writes changes correctly.

achimmihca:
Thanks for the info.

Probably an easy to fix issue.
Needs to be fixed by some dev who is using Linux.

Eisbaer1832:
So I managed to find a workaround. 
You can add the _Favorites_ playlist manually in the playlist editor, which resolves both the menu and icon issue respectively, until you reopen the game. 

Interestingly although playlists are being saved correctly, they are not being recreated when you relaunch the game.

Since the icons depend on the Playlists(Manager?) being initialized, this likely causes the ui issues.
Commenting out line 528 of the SongSelectEntryControl script therefore fixes the display issues by disabling the favorite icon. This however doesn't change the broken playlist feature on Linux.

Tuupertunut:
**The root cause is here:**

https://github.com/UltraStar-Deluxe/Play/blob/2e4f8b12ea3ede29aba35abc79123976f3d62d6f/UltraStar%20Play/Packages/playshared/Runtime/FileScanner/FileScanner.cs#L15-L72

[...]

PR #574, opened 2026-03-25
Fix hidden folder checking on Unix-based OS

Tuupertunut:
### What does this PR do?

Fixes a bug where all paths that had a hidden folder as a parent were ignored, even if the whole software was inside a hidden folder. On Linux the default config is found inside `$HOME/.config` so the whole config folder was ignored, causing a wide range of issues. The fix is to only ignore hidden folders that are children of the path being searched.

As discussed in #538 it might still be wise to review usage of the file scanner, as many or maybe all usages may not actually need to ignore hidden files. However this PR fixes the acute issue.

### Closes Issue(s)

#538, #569, maybe also some other Linux and MacOS issues.

### DIFF
--- a/UltraStar Play/Packages/playshared/Runtime/FileScanner/FileScanner.cs
+++ b/UltraStar Play/Packages/playshared/Runtime/FileScanner/FileScanner.cs
@@ -52,7 +52,7 @@ public static class FileScanner
         List<string> filteredResult = unfilteredResult
             // Ignore hidden files and folders
             .Where(filePath => (!config.ExcludeHiddenFiles || !IsHiddenFile(filePath))
-                               && (!config.ExcludeHiddenFolders || !IsInsideHiddenFolder(filePath)))
+                               && (!config.ExcludeHiddenFolders || !IsInsideHiddenFolder(filePath, folder)))
             .ToList();
 
         return filteredResult;
@@ -64,10 +64,11 @@ public static class FileScanner
         return Path.GetFileName(filePath).StartsWith(".");
     }
 
-    private static bool IsInsideHiddenFolder(string filePath)
+    private static bool IsInsideHiddenFolder(string filePath, string rootPath)
     {
         // On Unix based operating systems, files and folders that start with a dot are hidden.
-        string folderPath = Path.GetDirectoryName(filePath);
-        return folderPath.Contains("\\.") || folderPath.Contains("/.");
+        // Only check for hidden folders that are children of the root path, not the root path itself or its parents.
+        string folderPath = Path.GetDirectoryName(filePath).Substring(rootPath.Length);
+        return folderPath.StartsWith(".") || folderPath.Contains("\\.") || folderPath.Contains("/.");
     }
 }
--- a/UltraStar Play/Packages/playshared/Runtime/FileScanner/FileScanner.cs
+++ b/UltraStar Play/Packages/playshared/Runtime/FileScanner/FileScanner.cs
@@ -67,8 +67,12 @@ public static class FileScanner
     private static bool IsInsideHiddenFolder(string filePath, string rootPath)
     {
         // On Unix based operating systems, files and folders that start with a dot are hidden.
-        // Only check for hidden folders that are children of the root path, not the root path itself or its parents.
-        string folderPath = Path.GetDirectoryName(filePath).Substring(rootPath.Length);
+        // Only check for hidden folders that are children of the root path, not the root path
+        // itself or its parents.
+        // We know rootPath is always a prefix of filePath, so it can be cut out with Substring().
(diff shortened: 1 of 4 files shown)


Tuupertunut:
What should I do with the AI review? It may point out potential bugs but it may also very well hallucinate some imaginary problems. The changes it suggest seem overly complicated. Should I answer it or what?

UltraStar Play/Packages/playshared/Runtime/FileScanner/FileScanner.cs
```diff
@@ -64,10 +64,11 @@ private static bool IsHiddenFile(string filePath)
...
     }
 
-    private static bool IsInsideHiddenFolder(string filePath)
+    private static bool IsInsideHiddenFolder(string filePath, string rootPath)
     {
         // On Unix based operating systems, files and folders that start with a dot are hidden.
-        string folderPath = Path.GetDirectoryName(filePath);
-        return folderPath.Contains("\\.") || folderPath.Contains("/.");
+        // Only check for hidden folders that are children of the root path, not the root path itself or its parents.
+        string folderPath = Path.GetDirectoryName(filePath).Substring(rootPath.Length);
```
achimmihca: Thank you for tackling this.
I think this should work fine when rootPath is known to be a prefix of filePath.
But I am not sure whether this is the case when only looking at this code.

Consider to wrap rootPath in `Path.GetFullPath()`, to make it obvious that both are absolute paths.

Besides, performance is relevant here because there can be thousands of files (some users have +20k UltraStar songs). So only wrap it once, not inside the loop.

Please test performance with lots of dummy files. Ideally, performance should be similar as before. For example, this repo has several thousand txt files that you can use to test performance https://github.com/Asvarox/allkaraoke

achimmihca: > AI review [...] The changes it suggest seem overly complicated.

I agree, seems over engineered. Also performance has to be tested.

Tuupertunut: I'm pretty sure `rootPath` will always be a prefix of `filePath`. The [Directory.GetFiles documentation](https://learn.microsoft.com/en-us/dotnet/api/system.io.directory.getfiles?view=net-8.0) says that

> The returned file names are appended to the supplied path parameter.

I read that as a promise that the path parameter is always a prefix of any returned file paths. A quick test confirms that if you give it `.` it will return `./a` and `./b`. If you give it `/home/user/dir` it will return `/home/user/dir/a` and `/home/user/dir/b`. If you give it `../..` it will return `../../a` and `../../b`.

I agree it's not clear from the code that this is the case, so maybe a comment would clarify it. Performance should be about the same as before as `Substring` and `StartsWith` are cheap operations compared to filesystem checks. I considered adding an explicit check if `rootPath` is a prefix of `filePath` before doing `Substring` but that feels unnecessary as the documentation seems to already promise that.

Tuupertunut: I added a comment for that, as well as fixed a bug where `Path.GetDirectoryName()` removed trailing or duplicate slashes, resulting in changes in the string length. I also added a test case for that and the tests pass.

Tuupertunut:
I added edit mode tests for the original problem and noticed the same issue was also in `DirectoryScanner` and fixed it there as well. The file scanner code should now be quite robust.

There is a third place where hidden folders are checked: in `SongFolderListEntryControl.PathContainsHiddenFolder()`. I don't know how that's supposed to work so I don't know if that needs fixing, but it seems to only print warnings anyway.

achimmihca:
Thank you! I plan to review this and the other PRs next week.

UltraStar Play/Packages/playshared/Runtime/FileScanner/FileScanner.cs
```diff
@@ -64,10 +64,15 @@ private static bool IsHiddenFile(string filePath)
         return Path.GetFileName(filePath).StartsWith(".");
     }
 
-    private static bool IsInsideHiddenFolder(string filePath)
+    private static bool IsInsideHiddenFolder(string filePath, string rootPath)
```
achimmihca, line 67: Would be nice to have a common method for this. Currently, it is duplicated across DirectoryScanner and FileScanner.

Consider a HiddenFolderChecker or similar.

achimmihca (approved):

UltraStar Play/Packages/playshared/Runtime/FileScanner/FileScanner.cs
Tuupertunut: FileScanner and DirectoryScanner seem to be duplicates of each other anyway, they could probably be merged.

Issue #552, opened 2025-09-12
Space inside note should be allowed


zheka21:
Issue type: Development discussion

### Problem or question to be solved
In editor mode / lyrics editor not allowed to have space inside note.
It never happens in english but In russian (and not only in russian) language, in every single song we have prepositions "К тебе", "В этом", "С тобой" e.t.c. that should be treated as one note that has space inside "К те"-"бе", "В э"-"том", "С то"-бой" (it pronounced as "Кте"-"бе", "Вэ"-"том", "Сто"-бой")
If you open any russian ultrastar song in UltraStar-Deluxe/Play editor and click on edit lyrics it will be broken (empty notes, missing lyrics, missing letters)

### Suggested solution/s
remove function where space automatically split note

### Your opinion on the topic
Your editor is far more advanced and user friendlier than standard ultrastar deluxe editor but this bug makes it useless

achimmihca:
Hi, thank you for pointing this out. I was not aware of this.

> remove function where space automatically split note

Will make this optional in the next release.

As a workaround, I think it would work if you use a [non-breaking space](https://en.wikipedia.org/wiki/Non-breaking_space) because this should be a different character than regular space. How to enter a NBSP depends on the operating system.

zheka21:
Thank you for quick response. Will wait on new release.
Off topic. 
It would be great to implement customizable themes (black background screen in latest 1.5 build looks weird) and custom user avatars creation. 
Not sure if it is possible to store all config info in separate editable file so it can be copied/transferred to a new build. I know that Unity stores variables (Prefs) in some special places on your PC and they interfere if you have more than one version. Every time you run previous version it changes setting in a new one (at least "Graphics" and "Song Directory" settings) which is annoying.

Tuupertunut:
I think a solution could be to add an escape character in the lyrics editor like `\` so that `\ ` would be a space without line break, `\;` would be a semicolon without line break and `\\` would be a literal backslash. This would still work the same old way for most songs and also for most words in slavic languages but would support these characters when necessary.

achimmihca:
Merged in USPlay and Melody Mania 1.6.1-beta

PR #575, opened 2026-04-18
Implement escape characters for lyrics and lyrics code refactor

Tuupertunut:
### What does this PR do?

As mentioned in #552, the current lyrics editor does not support spaces inside syllables, because space has a special meaning as a word separator. This PR implements escape characters for the lyrics editor, so that you can write `К\ те` instead of `К те` if you want a literal space character inside the syllable. Because `;` is also a special syllable separator and adding `\` as an escape character makes it a special character too, `;` and `\` can also be escaped. The full list of escape sequences are therefore:

- `\ ` becomes ` `
- `\;` becomes `;`
- `\\` becomes `\`

Newline `\n` has no escape sequence because it can't be escaped. Word and syllable separators are special characters in the US Play editor but newline is a special character in the UltraStar TXT format and it can't appear in lyrics.

While implementing this feature, I noticed that the same escape processing logic had to be repeated in at least 5 different locations in the code, as you can see in the commit d46842d. I refactored the lyrics editing code so that all lyrics text processing now happens in a common file `LyricsUtils.cs` which has common utility functions for parsing and formatting lyrics text (commit eb98813). This utility class is then used by the big lyrics text field, the sentence level and note level text fields as well as the lyrics importer and hyphenation codes.

As usual when refactoring code, a number of small bugs were also found and fixed, such as

- Note splitting when editing a note with a leading space character.
- Whitespace replacement characters appearing temporarily in the real lyrics when editing a note.
- Not possible to add newlines to the lyrics field by pressing enter.

More information in commit messages and code comments.

### Closes Issue(s)

#552

### More

### Additional Notes

- I have understood that in Melody Mania 1.5.2 there is a new option to change the word separator character. I don't have access to Melody Mania 1.5.2 and I don't know how that's implemented, but these changes will probably conflict with that one. At least the code in this PR assumes that the real word separator and the one that the user writes on their keyboard is the same. If the user changes their word separator to pipe `|` they probably don't want the pipe to end up in their song files. Now that spaces can be escaped, I don't know if there's really a need for that option anymore. Please tell me if this feature still needs to be implemented.

- These changes were thoroughly tested by editing lyrics in the big lyrics text field, sentence level fields and note level fields, as well as loading song files with special characters. They were however not tested with an imported LRC file or the speech recognition feature but just assumed to work as before.


achimmihca:
Thank you for this PR. I think it is a useful feature.
Will take a look and merge when I find time.

I only have few questions:
- did you run the existing song editor tests, to check that other stuff still works?
- could you add a new Play Mode test for this feature (see `SongEditorPlayModeTests` folder)? For example
  - open song in song editor (see existing tests)
  - change text in lyrics area
  - assert that text on notes is as expected

I will try to resolve merge conflicts with MM in the next integrate. Don't worry about that.

Tuupertunut:
The existing tests pass. Play mode tests seem to be very clumsy to write so I'm not sure if I want to write any.

achimmihca:
> Play mode tests seem to be very clumsy to write so I'm not sure if I want to write any.

Ok, I will give it a try.
I agree, the test setup could need an overhaul.

Tuupertunut:
Maybe edit mode tests would be useful for the LyricsUtils.

achimmihca:
I will try to integrate this change in USPlay, then Melody Mania, and then merge back Melody Mania into USPlay.

This should align existing tests and code base.
But it will take me a while. I am busy right now with other stuff.

UltraStar Play/Assets/Common/Utils/LyricsUtils.cs
```diff
@@ -0,0 +1,237 @@
...
+    public static string GetEditModeText(Note note)
+    {
+        return FormatAsEditable(new[] { note.Text });
+    }
+
+    /// <summary>
+    /// Splits the text into sentences and syllables and applies it to the notes of the voice,
+    /// padding with empty strings if necessary.
+    /// </summary>
+    /// <param name="editModeText"></param>
```
achimmihca, line 52: I prefer JavaDoc resp. JsDoc format, because I think the XML stuff is rather verbose. Especially because it repeats `<summary>` always.

Furthermore, empty parameter descriptions are not helpful and tend to diverge with code over time. Please remove. The empty ones.

achimmihca (approved):

UltraStar Play/Assets/Common/Utils/LyricsUtils.cs
achimmihca: Anyway, overall this works well and I do not want to be nitpicky with the comments.
Still in the future, prefer to use JavaDoc (I know this is not very C#-ish) and avoid empty parameter descriptions.

